PR #7, opened 2020-02-21
Potential deadlock fix

todi1856:


vincent-unity:
Merged anyway.
Seems this branch didn't have the latest changes on master for yamato, so the check will never pass...

PR #8, opened 2020-02-23
Expose java source

vincent-unity:
This pr mainly exposes java source code as a part of the package, we don't need to build the .aar plugin, things that are worth mentioning:
1. The java source code is put under the `Runtime\Android\Plugins\Android\` folder, which is consistent to iOS.
2. `AndroidNotificationResourcesPostProcessor .InsertGradleDependencies()` function is added to insert the `com.android.support:appcompat-v7:27.1.1` dependency that needs by the java code.
3. `AndroidNotificationResourcesPostProcessor.InjectAndroidManifest` function is added to inject the `receiver`s to the manifest.
4. Did some code cleanup to `AndroidNotificationResourcesPostProcessor`.

Next:
1. Remove the .aar build code from yamato, will do it after this pr.
2. Add tests for gradle & manifest injections.
2. I have to comment `BuildConfig` references in the java code at https://github.com/Unity-Technologies/com.unity.mobile.notifications/commit/ce6e9023290e854d915424c6d8f356581b421b74. 
The reason is in 2019.3, the unity code and our jave source code will be included in `unityLibary` module which always has the `BuildConfig.java` file in `com.unity3d.player` package. 
But this is not true for 2018.4, which has the `BuildConfig.java` file in the package with the same name as the apk package...
Not sure what is the right solution to solve this issue, a potential solution is injecting the import code in the post processor as we do for gradle and manifest. But it's kind of nasty to me...

Issues about making java code public:
1. As you can see we have to do a lot of post processing in the callback, which is not very nice.
2. Fixing the BuildConfig might introduce worse code...
3. ~Developing the java code is not convenient, as you can't directly modify it. You have to export to gradle project to use android stuido as the IDE, to ensure the code still compile, then copy your changes back to the original java code...~
We can use `symbolink resources` while exporting to gradle project in build settings to solve this issue. It's not available for 2018.4, but probably good enough for us.


com.unity.mobile.notifications/Editor/AndroidNotificationResourcesPostProcessor.cs
```diff
@@ -1,142 +1,216 @@
+#if UNITY_EDITOR && PLATFORM_ANDROID
```
todi1856: Could we change PLATFORM_ANDROID->UNITY_ANDROID
since PLATFORM_ANDROID is not documented

com.unity.mobile.notifications/Editor/AndroidNotificationResourcesPostProcessor.cs
```diff
@@ -1,142 +1,216 @@
...
+        // Insert dependencies that need by mobile notification package.
+        private void InsertGradleDependencies(string projectPath)
+        {
+            // Here always insert a '\n' at the beginning, as in gradle:
+            //  1. for dependencies, you can put '}' at the end of the last 'implementation' line;
+            //  2. but you can't put two 'implementation's in one line.
+            // so just always add a new line to make sure it work for all cases.
+            const string kDependency = "\n    implementation 'com.android.support:appcompat-v7:27.1.1'\n";
+
+            var gradleFilePath = Path.Combine(projectPath, "build.gradle");
```
todi1856: shouldn't we throw exception instead silently failing? Please throw exceptions everywhere where you do return

com.unity.mobile.notifications/Editor/AndroidNotificationResourcesPostProcessor.cs
```diff
@@ -1,142 +1,216 @@
...
+
+            var content = File.ReadAllText(gradleFilePath);
+            if (string.IsNullOrEmpty(content))
+                return;
+
+            // Find the first '}' after 'dependencies' which has 'implementation' come after.
+            var regex = new Regex(@"dependencies[\s\S]+?implementation[\s\S]+?(?<index>}+?)");
+            var result = regex.Match(content);
+            if (result.Success)
+                File.WriteAllText(gradleFilePath, content.Insert(result.Groups["index"].Index, kDependency));
```
todi1856: exception for failing regex

com.unity.mobile.notifications/Editor/AndroidNotificationResourcesPostProcessor.cs
```diff
@@ -1,142 +1,216 @@
...
+                    File.WriteAllBytes(fileInfo.FullName, icon.Value);
+                }
+            }
+        }
+
+        private void InjectAndroidManifest(string projectPath)
+        {
+            var manifestPath = string.Format("{0}/src/main/AndroidManifest.xml", projectPath);
+            if (!File.Exists(manifestPath))
+                return;
```
todi1856: exception please

com.unity.mobile.notifications/Editor/AndroidNotificationResourcesPostProcessor.cs
vincent-unity: We can, I just didn't touch that... 
And we can remove UNITY_EDITOR as it's under Editor folder already.

com.unity.mobile.notifications/Runtime/Android/Plugins/Android/com/unity/androidnotifications/UnityNotificationRestartOnBootReceiver.java
```diff
@@ -0,0 +1,55 @@
...
+                long fireTime = data_intent.getLongExtra("fireTime", 0L);
+                Date currentDate = Calendar.getInstance().getTime();
+                Date fireTimeDate=new Date(fireTime);
+
+                int id = data_intent.getIntExtra("id", -1);
+                boolean isRepeatable = data_intent.getLongExtra("repeatInterval", 0L) > 0;
+
+                if (fireTimeDate.after(currentDate) || isRepeatable) {
+
+                    // if (BuildConfig.DEBUG) {
```
todi1856: Not sure, but there is this com.unity3d.player.CPPConstants class, which has UNITY_DEVELOPMENT_PLAYER property, and luckily it's public, maybe you could use that. I imagine if we build to Android dev player, I would want these logs, while for Non Development, I don't

com.unity.mobile.notifications/Editor/AndroidNotificationResourcesPostProcessor.cs
vincent-unity: Fxied at https://github.com/Unity-Technologies/com.unity.mobile.notifications/pull/8/commits/508ace7521feba484ac46d15408b7810ebe73029

com.unity.mobile.notifications/Editor/AndroidNotificationResourcesPostProcessor.cs
vincent-unity: Fixed at https://github.com/Unity-Technologies/com.unity.mobile.notifications/pull/8/commits/508ace7521feba484ac46d15408b7810ebe73029

com.unity.mobile.notifications/Editor/AndroidNotificationResourcesPostProcessor.cs
vincent-unity: Fixed at https://github.com/Unity-Technologies/com.unity.mobile.notifications/pull/8/commits/508ace7521feba484ac46d15408b7810ebe73029

com.unity.mobile.notifications/Editor/AndroidNotificationResourcesPostProcessor.cs
vincent-unity: Fixed at https://github.com/Unity-Technologies/com.unity.mobile.notifications/pull/8/commits/508ace7521feba484ac46d15408b7810ebe73029

com.unity.mobile.notifications/Runtime/Android/Plugins/Android/com/unity/androidnotifications/UnityNotificationRestartOnBootReceiver.java
vincent-unity: Talked with Tomas offline about this.
1. com.unity3d.player.CPPConstants is private which can't be used here.
2. Once alternative is generating similar build config by our own in the post processor, which probably work for most of the cases. 
But this won't work for android studio, as these are two levels build config. 
3. The idea way is still making it work with BuildConfig.java, but I can't think of a way to do that without injecting the `import` code to the jave sources.

As a conclusion, we probably will ignore it for now.

PR #11, opened 2020-03-09
Remove the unncessary `iOS` directory from the plugin path.

vincent-unity:
Remove the duplicate `iOS` in the path. 
Will do Android in a separate pr with refactoring.

To QA: please simply verify if the test project works on iOS. It worked for me on my iPad.


vincent-unity:
Pushed a new version to remove the `notifications` folder from path.

PR #12, opened 2020-03-10
Android plugin code refactoring

vincent-unity:
This pr mainly did the Android plugin code refactoring, including:
1. Reorder the functions in `UnityNotificationManager` by grouping them with functionalities.
2. Only make APIs public when necessary, change others to protected if it's only used internally.
3. Removed `UnityNotificationManager.getScheduledNotifications()` and its related functions as they're not used for now, we can add them back when we really want to use.
4. Move some APIs to `UnityNotificationManagerUtilities`, like `serializeNotificationIntent/deserializeNotificationIntent` etc.
5. Rename some functions to java style, also format the files to follow java convention.
6. Add simple comment to the funtions, also some code snippets.
7. Use `name` rather than `title` for NotificationChannel, as `name` is what Android is using. 
8. Other minor changes

To QA: Please help to verify if it on 2018.4/2019.3 as well, also please verify on device lower than Android N, as I don't have any Android 6 or below devices.

PR #13, opened 2020-03-11
Notification nougat refactoring

vincent-unity:
This pr mainly contains some discussion and comments from https://github.com/Unity-Technologies/com.unity.mobile.notifications/pull/12. Just to make the changes more trackable by doing in a separate pr.

1. Remove `UnityNotificationManagerNougat` as we don't need to do something special for Android N.
2. Change version check to assert in `UnityNotificationManagerOreo`.
3. Avoid redundant string array construction.

To QA: Please also test on devices lower than Android N. Thanks.

P.S. Please don't merge as I just target on `android-plugin-refactoring` branch which is not in master yet.


vincent-unity:
@lucianUnity3D FYI.

vincent-unity:
> However, Xiaomi devices seem to no support some of the features, such as .color property and custom icons.
Thanks Vaidas. Do you know if it's a regression or has been like that always?

PR #14, opened 2020-03-12
Use GetBuildPropertyForAnyConfig() as we're using main target guid.

vincent-unity:
Fixed case https://fogbugz.unity3d.com/f/cases/1221133/ by using GetBuildPropertyForAnyConfig() as main target guid is used.

To QA: Please check there're only one entitlements files generated, both for 2018.4 & 2019.3. Thanks.


vincent-unity:
@lucianUnity3D FYI.

PR #20, opened 2020-03-20
Fix qa round comments

vincent-unity:
Fixed several issues that are found during qa pass.
1. Throw exceptions for invalid argument like 0 time interval, less than 60s for repeating notifications, rather than doing nothing or just a debug warning.
2. Have to add #pragma warning 162 back for now, will find a better way to do it along with iOS plugin code polishing.

### DIFF
--- a/com.unity.mobile.notifications/Runtime/iOS/iOSNotification.cs
+++ b/com.unity.mobile.notifications/Runtime/iOS/iOSNotification.cs
@@ -199,17 +199,10 @@ namespace Unity.Notifications.iOS
                     data.triggerType = iOSNotificationTimeIntervalTrigger.Type;
                     data.timeTriggerInterval = trigger.timeInterval;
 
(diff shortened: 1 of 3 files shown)


com.unity.mobile.notifications/Runtime/iOS/iOSNotification.cs
```diff
@@ -199,17 +199,10 @@ public iOSNotificationTrigger Trigger
...
-                    }
-                    else
-                    {
-                        if (trigger.Repeats)
-                        {
-                            Debug.LogWarning("Time interval must be at least 60 for repeating notifications.");
-                        }
-                    }
+                    if (trigger.Repeats && trigger.timeInterval < 60)
+                        throw new ArgumentException("Time interval must be 60 seconds or greater for repeating notifications.");
```
todi1856: would be nice to add the value of  trigger.timeInterval in the exception

PR #23, opened 2020-03-25
Notification settings refactoring (Renaming files and classes)

vincent-unity:
Renaming files and classes.


com.unity.mobile.notifications/Editor/NotificationSetting.cs
```diff
@@ -2,19 +2,19 @@
...
+    internal class NotificationSetting
     {
         public string key;
         public string label;
         public string tooltip;
         public object val;
         public bool writeToPlist;
 
-        public List<NotificationEditorSetting> dependentSettings;
+        public List<NotificationSetting> dependentSettings;
```
todi1856: For latter, public viarables have to be upper case

PR #25, opened 2020-03-25
Updating notification samples project

renaldasrep:
Replacing the old sample project with a new one that is more convenient to use as it has a scrollable list of buttons and a text field that logs out the methods used and their status

Multiple scenes are no longer needed, iOS and Android scripts are loaded on start depending on which platform the project is running on

image

It includes multiple manfest.json files appended with major Unity versions, so that you could switch the real manifest.json contents and open the project on different versions without breaking it

image

It also uses Scriptable Objects to more conveniently create notification templates that can then be scheduled through the UI buttons

image


TestProjects/NotificationSamples/Assets/Scripts/Helpers/GameObjectReferences.cs
```diff
@@ -0,0 +1,10 @@
+﻿using UnityEngine;
+using UnityEngine.UI;
+
+public class GameObjectReferences : MonoBehaviour
+{
+    public Text logsText;
+    public Transform buttonScrollViewContent;
+    public Transform buttonGroupTemplate;
+    public ScrollRect logsScrollRect;
```
vincent-unity: Public fields should start with Capital according to our coding covention.

TestProjects/NotificationSamples/Assets/Scripts/AndroidTest.cs
```diff
@@ -0,0 +1,262 @@
...
+
+public class AndroidTest : MonoBehaviour
+{
+    public List<AndroidNotificationTemplate> androidNotifications;
+
+#if PLATFORM_ANDROID
+
+    private GameObjectReferences gameObjectReferences;
+    private Dictionary<string, OrderedDictionary> groups;
+    private Logger LOGGER;
```
vincent-unity: Private fields should start with `m_`.

vincent-unity:
Thanks for improving the test project, it's much better to use. Few things in my mind:
1. One concern to me is if we really need package manifest for every versions. IIRC, only 2018.x is not compatible with 2019.x and above. Probably we can only keep two versions for 2018.4 and 2019.3?
2. Can we output the device token on iOS while requesting authorization? It will be convenient for sending remote notification.
3. You have to run the formatter, check `format.ps1` or `format.sh`  under the root folder.

vincent-unity:
Can we make the text a little bigger, for the sake of devices with no that big screen? :)
I tried to run the sample on Google pixel 2, it's really hard to see the text...

TestProjects/NotificationSamples/Assets/Scripts/AndroidTest.cs
```diff
@@ -0,0 +1,262 @@
...
+
+#if PLATFORM_ANDROID
+using Unity.Notifications.Android;
+#endif
+
+public class AndroidTest : MonoBehaviour
+{
+    public List<AndroidNotificationTemplate> androidNotifications;
+
+#if PLATFORM_ANDROID
```
todi1856: wondering, why are you wrapping this code with Android, this project is referencing notification packages, so all the API you use here, is also available in Editor. So why wrap it?

TestProjects/NotificationSamples/Assets/Scripts/Helpers/PlatformSelector.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+using UnityEngine;
+
+public class PlatformSelector : MonoBehaviour
+{
+    void Awake()
+    {
+        #if PLATFORM_ANDROID
```
todi1856: I think you should actually use a swtich with  Application.platform instead of defines, that why you'll get compile errors if there are any in advance

TestProjects/NotificationSamples/Assets/Scripts/ScriptableObjects/AndroidNotificationTemplate.cs
```diff
@@ -0,0 +1,40 @@
...
+using UnityEngine;
+
+#if PLATFORM_ANDROID
+using Unity.Notifications.Android;
+#endif
+
+[CreateAssetMenu(menuName = "Mobile Notifications/Android Notification Template")]
+public class AndroidNotificationTemplate : ScriptableObject
+{
+    #if PLATFORM_ANDROID
```
todi1856: No need for this define, let the code run in Editor too

TestProjects/NotificationSamples/Assets/Scripts/Helpers/PlatformSelector.cs
renaldasrep: Done in latest commit

TestProjects/NotificationSamples/Assets/Scripts/ScriptableObjects/AndroidNotificationTemplate.cs
renaldasrep: I added "|| UNITY_EDITOR" to these ifdefs in latest commit, because it doesn't compile if I remove them completely - e.g. when building on Android it tries to compile files that reference iOS namespace

TestProjects/NotificationSamples/Assets/Scripts/AndroidTest.cs
renaldasrep: I added "|| UNITY_EDITOR" to these ifdefs in latest commit, because it doesn't compile if I remove them completely - e.g. when building on Android it tries to compile files that reference iOS namespace

TestProjects/NotificationSamples/Assets/Scripts/Helpers/GameObjectReferences.cs
renaldasrep: fixed these, and some others I found, hope I found them all

TestProjects/NotificationSamples/Assets/Scripts/AndroidTest.cs
renaldasrep: fixed these, and some others I found, hope I found them all

renaldasrep:
@vincent-unity 

> 1. One concern to me is if we really need package manifest for every versions. IIRC, only 2018.x is not compatible with 2019.x and above. Probably we can only keep two versions for 2018.4 and 2019.3?

I added a comment to your PR, there's some considerations regarding 2019.1

> 2. Can we output the device token on iOS while requesting authorization? It will be convenient for sending remote notification.

I already Debug.Log it so it's easy to copy and use outside the player Do you think it's necessary to output it to in-player log?

> 3. You have to run the formatter, check format.ps1 or format.sh under the root folder.

Done! :)

renaldasrep:
@vincent-unity 

>Can we make the text bigger, for the sake of devices with no that big screen? :)
I tried to run the sample on Google pixel 2, it's really hard to see the text...
On iPad, it looks good.

Added some adjustments to canvas scaling in today's commits, not text & buttons are definitely bigger on a Pixel 2, I also added new fonts (a condensed one to fit more characters in a single line of button, and a different monospace one (grabbed them off Google Fonts, as far as I understand they're free to use)

.gitignore
```diff
@@ -17,3 +17,4 @@ android-sdk/*
 *.suo
 *.sublime-project
 *.sublime-workspace
+*.preformat.bak
```
vincent-unity: Please don't do that. Personally I don't like this, the reason is if you add this kind of files into gitingore, you don't have a center place to remove all of them in one batch.
For example, if there're several .bak files show in Unity and scatter in different folders, you have to go to different folders to delete all of them, it's a bit annoying to me.
For now, they all show in git (SourceTree for instance), I can delete all of them in one batch...

Wondering if the formatter tool has the option to go bold without generating the .bak files...

TestProjects/NotificationSamples/Assets/Scripts/ScriptableObjects/AndroidNotificationTemplate.cs
vincent-unity: The reason is we only have Editor & iOS checked for Unity.Notification.iOS in the .asmdef. The same to Android.
So either we add #ifdef or we add different .asmdef for the testing script.

.gitignore
renaldasrep: updated the format script to clean up after itself 👍

vincent-unity: 👍

PR #29, opened 2020-03-30
Internal test packge

vincent-unity:
Move all the content in the sample to an internal package which can be shared by multiple versions of testing sample projects.
Now the folder structure is

TestProjects
--- com.unity.mobile-notifications-sample
--- NotificationSamples_2018_4
--- NotificationSamples_2019_3

I only tested 2019.3 on Google pixel 2, worked fine for me.


vincent-unity:
> opening on 2019.1 is still finicky
> 
> You either have to remove vscode, rider, androidjni packages from 2019.3 project and open that
> OR open 2018.3 project and remove packages that throw errors (because they are added automatically, and they're not in manifest originally)
> 
> I have to approve because everything else seems to be working and I can't come up with good ways to solve 2019.1 issue

We always can add a project for 2019.1 version, as long as it's the version we use in daily work...

PR #31, opened 2020-04-01
Fix incorrect string format in AndroidNotificationCenter

ondrowan:


com.unity.mobile.notifications/Runtime/Android/AndroidNotificationCenter.cs
```diff
@@ -130,11 +130,11 @@ public static void RegisterNotificationChannel(AndroidNotificationChannel channe
             }
             else if (string.IsNullOrEmpty(channel.id))
```
vincent-unity, line 131: Here should check `channel.Name`... 
Would you like to fix it at your side or let me fix from the main repo?

vincent-unity: I'll merge anyway. Thanks.

PR #36, opened 2020-04-13
Notification settings ui improvement

vincent-unity:
Improved the notification settings UI, including
1. Moving the icon requirement message below the `Notification Icons` header.
2. Improve the drawing code a little bit.
3. Fixed the issue that dependent settings are not aligned correctly.
4. Fixed https://fogbugz.unity3d.com/f/cases/1195293/, now scrollview works fine with notification settings.
5. Other minor changes.

Now the UI looks like
image

PR #41, opened 2020-04-16
Bump version to 1.2.0-preview.

vincent-unity:
Prepare a new 1.2.0 release:
1. Improved notification setting UI change
2. Moved the notification setting asset under 'Project Settings'
3. Fixed some bugs on iOS and Android.

QA: Please verify the il2cpp build on Android as we removed the long[] to int[] workaround for vibration patterns.

### DIFF
--- a/com.unity.mobile.notifications/CHANGELOG.md
+++ b/com.unity.mobile.notifications/CHANGELOG.md
@@ -2,10 +2,20 @@
 
 All notable changes to this package will be documented in this file.
 
+## [1.2.0-preview] - 2020-04-16
+
+### Changes & Improvements:
+- [Editor] Improved the Notification Settings UI, also stored the notification setting asset under 'Project Settings' folder rather than 'Assets' folder.
+
+### Fixes:
+- Fixed a lot of notification settings UI related bugs.
+- [iOS] Fixed the issue that the authorization request callback hasn't been removed.
+- [Android] Fixed the wrong value of LockscreenVisiblity.Private.
+
 ## [1.1.0-preview] - 2020-03-24
 
 ### Changes & Improvements:
-- [Android] Exposing the java source code in the package rather than building into an .aar. Users can now debug the java code in Android Studio by exporting to a gradle project.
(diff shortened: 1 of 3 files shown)


vincent-unity:
@lucianUnity3D FYI.

com.unity.mobile.notifications/CHANGELOG.md
```diff
@@ -2,10 +2,20 @@
 
 All notable changes to this package will be documented in this file.
 
+## [1.2.0-preview] - 2020-04-16
+
+### Changes & Improvements:
+- [Editor] Improved the Notification Settings UI, also stored the notification setting asset under 'Project Settings' folder rather than 'Assets' folder.
```
todi1856, line 8: It's 'ProjectSettings', not 'Project Settings' :)

vincent-unity: Nice catch. Fixed.

vaidasma:
Works as intended on IL2CPP and Mono (.NET 4.x and .NET Standard 2.0).
The only concern that I have from the UX side of the project settings is that the Notification window doesn't scale well when shrinking the window horizontally (check the right side of the window).  @vincent-unity you think you can implent an easy fix for this? 
image

vincent-unity:
@vaidasma I think the size is limited with minimal 640 at the very beginning, so I didn't change that part but only made the scrollbar work correctly. If you check other settings like Editor & Player, you can also see the scrollbar at the bottom but with a smaller minimal value.

It's not an easy fix but I can do it with https://github.com/Unity-Technologies/com.unity.mobile.notifications/pull/36#discussion_r409251293 in next pr, by changing to a smaller minimal value. How do you think?

vaidasma:
@vincent-unity Okay, that works. I can see other windows, such as Player settings, dealing better with resizing the default layout horizontally by shrinking some of the containers appropriately. Should I make a ticket so it would be easier to remember?

vincent-unity:
@vaidasma Please make a ticket. And I would also like to discuss with UX guys about this to see what the best behavior could be here.

Issue #44, opened 2020-04-25
Incompatible with GooglePlayResolver (External Dependency Manager)


newatlanta19006:
Because you now have .java files directly in this package. External Dependency Manager will not detect duplicate uses and the build will have issues due to conflicting api functions. 

The Resolver is pretty much essential in any project so support should be prioritized. 

I dont think there is a work around currently that I know of so I am downgrading to a previous version.

jounisuom:
I also have this issue because the AndroidNotificationPostProcessor adds the com.android.support:appcompat-v7:27.1.1 dependency to the project. Namely I get this error:

`
java.lang.RuntimeException: Duplicate class android.support.v4.app.INotificationSideChannel found in modules classes.jar (:androidx.core.core-1.2.0:) and classes.jar (com.android.support:support-compat:27.1.1)
`

I can't get rid of it because the dependency is always added.

We are also forced to use older version.

I suggest the following fix:
- remove the aforementioned dependency from the post-processor
- in java files, refactor android.support.annotation.Keep to androidx.annotation.Keep

vincent-unity:
Fixed in the new patch release 1.2.1-preview.

PR #47, opened 2020-04-30
Bug fixes

vincent-unity:
https://fogbugz.unity3d.com/f/cases/1231218/
https://fogbugz.unity3d.com/f/cases/1239769/
https://github.com/Unity-Technologies/com.unity.mobile.notifications/issues/44

This mainly fixed the android build failure introduced by `com.android.support:appcompat-v7:27.1.1` dependency. We use the `annotation.Keep` from that library but but we don't really need now.

I guess the reason we first added it is users can use the deassembly tools to check the class/method names of our android plugin, since we expose the source, we don't need it any more.

This gets rid of some bugs by not injecting the gradle file.

To QA: **_Could you please priotirize this?_** As this blocks people from upgrading to the 1.1.0-preview and above. I planned to have a patch release with it.
And I personally tested it on pixel 2 with android 8.

### DIFF
--- a/com.unity.mobile.notifications/Editor/AndroidNotificationPostProcessor.cs
+++ b/com.unity.mobile.notifications/Editor/AndroidNotificationPostProcessor.cs
@@ -16,37 +16,9 @@ namespace Unity.Notifications
 
         public void OnPostGenerateGradleAndroidProject(string projectPath)
         {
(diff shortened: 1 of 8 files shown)


vincent-unity:
@lucianUnity3D FYI.

PR #54, opened 2020-05-28
Bug fixes

vincent-unity:
Bug fixes including:
1. Fixed the notification color not working issue, also fixed the wrong order of RGBA.
2. Case 1199310: Fix the issue that Mobile Notification icons are gray when project is built from batch mode with -nographics flag.

### DIFF
--- a/com.unity.mobile.notifications/Editor/TextureAssetUtils.cs
+++ b/com.unity.mobile.notifications/Editor/TextureAssetUtils.cs
@@ -1,6 +1,7 @@
 using System.Collections.Generic;
 using UnityEditor;
 using UnityEngine;
+using UnityEngine.Rendering;
 
 namespace Unity.Notifications
 {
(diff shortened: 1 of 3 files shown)


vincent-unity:
@lucianUnity3D FYI.

com.unity.mobile.notifications/Editor/TextureAssetUtils.cs
```diff
@@ -96,26 +97,46 @@ public static Texture2D ProcessTextureForType(Texture2D sourceTexture, Notificat
...
         {
-            if (sourceTexture.width == width && sourceTexture.height == sourceTexture.height)
+            if (sourceTexture.width == width && sourceTexture.height == height)
                 return sourceTexture;
 
-            Rect rect = new Rect(0, 0, width, height);
+            if (Application.isBatchMode && SystemInfo.graphicsDeviceType == GraphicsDeviceType.Null)
+            {
+                // If unity is running in batch mode without graphics, we have to go to the slow CPU code path.
+                var result = new Texture2D(width, height, TextureFormat.ARGB32, false);
```
todi1856, line 106: why not use always this path, since it seems like you might get inconsistent results, batch mode versus non batch mode

vincent-unity: The reason is gpu scaling should be faster... And it can do trilinear easily.

We probably can wait the Image class being exported to C# to have a consistent way to do the conversion.

vincent-unity: I did some profiling with the test project in the bug reports, seems like for small textures like the icon, the cpu way is pretty much fast the same level as gpu way. 

I'm testing with 2019.3.4f1 + two icons (256x256), 
1. the cpu path time: 20ms+
2. the gpu path time: 15ms+

Then I'll go with what you suggest by only keeping the cpu path.

vincent-unity:
@vaidasma Could you please verify if the `non` batch mode works fine? I personally verified it locally.

PR #56, opened 2020-06-02
Case 1244642: Fixed the bug that we can't really set the foreground p…

vincent-unity:
Fixed bug https://fogbugz.unity3d.com/f/cases/1244642/.
Previously we don't allow users to set presentation option to None, if you select Nothing, we converted it to All. This caused problem if users don't want to show anything for remote notifications if the app is running in foreground.

So I removed this restriction from both notification settings side and the native plugin side.

To QA:
I personally verify on iOS locally.


vaidasma:
Issue still present with Default Notification Authorization Options dropdown list

PR #64, opened 2020-09-08
Fix Android crash when many notifications are scheduled at once (case…

vaidas-unity:
Fix Android app crash when a high number of notifications is scheduled ([case 1172850](https://fogbugz.unity3d.com/f/cases/1172850/)).
When profiling the app, it was clear that the crash was caused by GC overhead caused by excess allocations and destructions each time a notification was polled. To remedy this, some of the iterative allocations were replaced by static, re-usable members of AndroidNotificationCenter class.

In the repro project, 1000 notifications are scheduled to cause a crash, but it seems that ~50 notifications is the maximum limit for the number of notifications the OS allows at once.

Tester notice:
* Please check the target FB case no longer reproduces.
* It may still be possible for the crash to occur on old/low-end devices, but the optimization in this PR reduces those chances greatly.

### DIFF
--- a/com.unity.mobile.notifications/Runtime/Android/AndroidNotificationCenter.cs
+++ b/com.unity.mobile.notifications/Runtime/Android/AndroidNotificationCenter.cs
@@ -43,7 +43,9 @@ namespace Unity.Notifications.Android
         /// </summary>
         public static event NotificationReceivedCallback OnNotificationReceived = delegate {};
 
+        private static AndroidJavaClass s_NotificationManagerClass;
         private static AndroidJavaObject s_NotificationManager;
+        private static AndroidJavaObject s_NotificationManagerContext;
         private static bool s_Initialized;
 
         public static bool Initialize()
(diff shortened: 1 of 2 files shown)


vincent-unity (approved):
Nice!

com.unity.mobile.notifications/Runtime/Android/AndroidNotificationCenter.cs
```diff
@@ -60,15 +62,17 @@ public static bool Initialize()
 #if UNITY_EDITOR || !UNITY_ANDROID
             s_NotificationManager = null;
             s_Initialized = false;
+            s_NotificationManagerClass = null;
+            s_NotificationManagerContext = null;
 #elif UNITY_ANDROID
             AndroidJavaClass unityPlayer = new AndroidJavaClass("com.unity3d.player.UnityPlayer");
             AndroidJavaObject activity = unityPlayer.GetStatic<AndroidJavaObject>("currentActivity");
```
vincent-unity, line 69: Do you think if we should do the same to these two variables(converting to s_***)? As we use them in GetLastNotificationIntent().

vaidas-unity: Good idea, thanks

vincent-unity (approved):

vaidasma:
Verified. Application no longer crashes. Notifications are sent and canceled over the period of 10-15 seconds (since it's 1000 notifications being processed)

vaidasma (approved):

PR #65, opened 2020-09-15
Update UnityNotificationManager.m

alexey-unity:
fixed crash on ios when using null title or body (case 1276575)


vincent-unity (approved):

vaidasma (approved):
Verified as fixed with iPhone 6s.

Application no longer crashes

PR #69, opened 2020-10-19
Location related fixes for iOS

aurimasc:
Fix iOSNotificationData struct marshalling (bools are marshalled as 32-bit integers, so far all worked fine thanks to alignment except for locationTriggerNotifyOnExit always being false (and would have issues if one more field was appended).
Location coordinate had typo and wasn't passed correctly.

### DIFF
--- a/com.unity.mobile.notifications/Runtime/iOS/Plugins/UnityNotificationData.h
+++ b/com.unity.mobile.notifications/Runtime/iOS/Plugins/UnityNotificationData.h
@@ -28,12 +28,12 @@ typedef struct iOSNotificationData
 
     //Custom data
     char* data;
-    BOOL showInForeground;
+    int showInForeground;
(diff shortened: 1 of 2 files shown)


com.unity.mobile.notifications/Runtime/iOS/Plugins/UnityNotificationData.h
```diff
@@ -28,12 +28,12 @@ typedef struct iOSNotificationData
 
     //Custom data
     char* data;
-    BOOL showInForeground;
+    int showInForeground;
     int showInForegroundPresentationOptions;
```
todi1856, line 33: P.S Something for the future, I think it's a good practice to pass struct size next to the marshalled struct, so you can validate that C# struct size matches native struct size

todi1856 (approved):

vaidasma (approved):

PR #75, opened 2020-11-16
UI improvements

aurimasc:
case 1195293
Notification Settings UI improvements:
- Label height & vertical width more consistent
- Checkbox placement adapted per platform (was made for iOS and way to wide for Android)
- Min left texture position on Android to not overlap with fields to it's left
- Vertical adjustments to couple messages


vaidas-unity (approved):
Looks good

vaidasma:
Looks good except one little detail: the text area does not handle the vertical overflow of the text field. It should either be scrollable or extend the bounds of the field. 

It's a small issue, however, I feel this one is quite visible for the users, therefore I feel like we should not present the users with flawed design.
image

vaidasma (approved):

PR #76, opened 2020-11-20
Thread-safety to AuthorizationRequest

aurimasc:
Reviewing commit-by-commit might be easier.

Mutiple fixes:
* AuthorizationRequest is filled in callback from other thread, need to sync
* Multiple requests can be submitted at once, need to handle that
  - the objc code before changes had race conditions if multiple requests at once
  - in C# we had a single static event, meaning one finishing request finishes all current ones
  - if request isn't disposed and another is created, the first one is also affected

### DIFF
--- a/com.unity.mobile.notifications/Runtime/iOS/AuthorizationRequest.cs
+++ b/com.unity.mobile.notifications/Runtime/iOS/AuthorizationRequest.cs
@@ -89,6 +89,11 @@ namespace Unity.Notifications.iOS
         /// </summary>
         public string DeviceToken { get; private set; }
 
+        static AuthorizationRequest()
+        {
+            iOSNotificationsWrapper.RegisterAuthorizationRequestCallback();
+        }
+
         /// <summary>
         /// Initiate an authorization request.
         /// </summary>
@@ -97,7 +102,6 @@ namespace Unity.Notifications.iOS
         /// If registration succeeds the DeviceToken will be returned. You should pass this token along to the server you use to generate remote notifications for the device. </param>
         public AuthorizationRequest(AuthorizationOption authorizationOption, bool registerForRemoteNotifications)
         {
-            iOSNotificationsWrapper.RegisterAuthorizationRequestCallback();
             iOSNotificationsWrapper.RequestAuthorization((int)authorizationOption, registerForRemoteNotifications);
 
             iOSNotificationCenter.OnAuthorizationRequestCompleted += OnAuthorizationRequestCompleted;
--- a/com.unity.mobile.notifications/Runtime/iOS/AuthorizationRequest.cs
+++ b/com.unity.mobile.notifications/Runtime/iOS/AuthorizationRequest.cs
@@ -147,8 +147,9 @@ namespace Unity.Notifications.iOS
                 handle.Free();
                 req.OnAuthorizationRequestCompleted(requestData);
             }
-            catch
+            catch (Exception e)
             {
+                Debug.LogException(e);
             }
         }
 
--- a/com.unity.mobile.notifications/Runtime/iOS/AuthorizationRequest.cs
+++ b/com.unity.mobile.notifications/Runtime/iOS/AuthorizationRequest.cs
@@ -34,9 +34,8 @@ namespace Unity.Notifications.iOS
     [StructLayout(LayoutKind.Sequential)]
     internal struct iOSAuthorizationRequestData
     {
-        internal bool granted;
+        internal int granted;
         internal string error;
-        internal bool finished;
         internal string deviceToken;
     }
 
@@ -132,8 +131,8 @@ namespace Unity.Notifications.iOS
         {
             lock (this)
             {
-                IsFinished = requestData.finished;
-                Granted = requestData.granted;
+                IsFinished = true;
+                Granted = requestData.granted != 0;
                 Error = requestData.error;
                 DeviceToken = requestData.deviceToken;
             }
--- a/com.unity.mobile.notifications/Runtime/iOS/AuthorizationRequest.cs
+++ b/com.unity.mobile.notifications/Runtime/iOS/AuthorizationRequest.cs
@@ -68,26 +68,48 @@ namespace Unity.Notifications.iOS
     /// </example>
     public class AuthorizationRequest : IDisposable
     {
+        bool _IsFinished;
+        bool _Granted;
+        string _Error;
+        string _DeviceToken;
+
         /// <summary>
         /// Indicates whether the authorization request has completed.
         /// </summary>
-        public bool IsFinished { get; private set; }
+        public bool IsFinished
+        {
+            get { lock (this) { return _IsFinished; } }
+            private set { _IsFinished = value; }
+        }
 
         /// <summary>
         /// A property indicating whether authorization was granted. The value of this parameter is set to true when authorization was granted for one or more options. The value is set to false when authorization is denied for all options.
         /// </summary>
-        public bool Granted { get; private set; }
+        public bool Granted
+        {
+            get { lock (this) { return _Granted; } }
+            private set { _Granted = value; }
+        }
 
         /// <summary>
         /// Contains error information of the request failed for some reason or an empty string if no error occurred.
         /// </summary>
-        public string Error { get; private set; }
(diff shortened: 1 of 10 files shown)


vaidas-unity (approved):

com.unity.mobile.notifications/Runtime/iOS/AuthorizationRequest.cs
```diff
@@ -68,26 +67,53 @@ internal struct iOSAuthorizationRequestData
     /// </example>
     public class AuthorizationRequest : IDisposable
```
vincent-unity, line 68: I didn't get why we need to lock `AuthorizationRequest` for getter...

com.unity.mobile.notifications/Runtime/iOS/AuthorizationRequest.cs
```diff
@@ -97,26 +123,40 @@ public class AuthorizationRequest : IDisposable
...
+        {
+            try
+            {
+                var handle = GCHandle.FromIntPtr(request);
+                var req = handle.Target as AuthorizationRequest;
+                handle.Free();
+                req.OnAuthorizationRequestCompleted(requestData);
+            }
+            catch
+            {
```
vincent-unity, line 151: An error log might be helpful to users.

vincent-unity (approved):

com.unity.mobile.notifications/Runtime/iOS/AuthorizationRequest.cs
aurimasc: We set properties from other thread, so we have a race condition between read&write, which might not be safe for strings. Also, you may have race condition if you read multiple properties, putting locks place ensures all properties are set atomically by the writer before they are made available to the reader.

com.unity.mobile.notifications/Runtime/iOS/AuthorizationRequest.cs
```diff
@@ -68,26 +67,53 @@ internal struct iOSAuthorizationRequestData
     /// </example>
     public class AuthorizationRequest : IDisposable
     {
+        bool _IsFinished;
```
todi1856, line 70: code style, private variable should be prefixed by m_

com.unity.mobile.notifications/Runtime/iOS/iOSNotificationsWrapper.cs
```diff
@@ -104,13 +102,13 @@ public static void RegisterOnReceivedCallback()
         }
 
         [MonoPInvokeCallback(typeof(AuthorizationRequestCallback))]
-        public static void AuthorizationRequestReceived(IntPtr authRequestDataPtr)
+        public static void AuthorizationRequestReceived(IntPtr request, IntPtr authRequestDataPtr)
```
todi1856, line 105: Since you're changing this code, can you make it even more safer. Add the incoming iOSAuthorizationRequestData  struct size, so you can then verify that incoming struct size matches struct size on C#, if it doesn't match throw an excption.

Also, I never understood why do we need Marshal.PtrToStructure ? and why not pass struct type.

So in short:
1. Change these types
private delegate void AuthorizationRequestCallback(IntPtr request, int size, iOSAuthorizationRequestData data);
public static void AuthorizationRequestReceived(IntPtr request,int size, iOSAuthorizationRequestData data)
2. Add validation for incoming objc iOSAuthorizationRequestData  size versus C# iOSAuthorizationRequestData size, if doesn't match throw an error, this will help avoid case where somebody updates or incorrectly adds type to struct
3. Remove Marshal.PtrToStructure and just use iOSAuthorizationRequestData directly, unless you know a reason why we should keep Marshal.PtrToStructure here

todi1856: Same for  private delegate void NotificationReceivedCallback(IntPtr notificationData); 

1. Add int size parameter before notificationData, so you can perform validation
2. And maybe instead IntPtr use iOSNotificationData directly, since the struct layout matches layout from objc.

com.unity.mobile.notifications/Runtime/iOS/AuthorizationRequest.cs
todi1856: What Vincent said, please log an error

todi1856 (changes requested):

com.unity.mobile.notifications/Runtime/iOS/iOSNotificationsWrapper.cs
todi1856: P.S if these changes will be done, please rerun QA pass, just in case

todi1856: still waiting for size parameter for validation, just in case

com.unity.mobile.notifications/Runtime/iOS/AuthorizationRequest.cs
aurimasc: Fixed.

com.unity.mobile.notifications/Runtime/iOS/AuthorizationRequest.cs
aurimasc: Renamed.

com.unity.mobile.notifications/Runtime/iOS/iOSNotificationsWrapper.cs
aurimasc: I don't like the idea of putting unrelated changes to this PR, so I'd like to resolve iOSNotificationData part in a separate PR.

As for iOSAuthorizationRequestData I took a bit different approach: a separate code that only runs once in development build. Lets not pollute production code with checks that must not fail.

todi1856: Can you log iOSNotificationData task somewhere so it wouldn't be forgotten?

com.unity.mobile.notifications/Runtime/iOS/iOSNotificationsWrapper.cs
```diff
@@ -70,20 +75,30 @@ internal class iOSNotificationsWrapper : MonoBehaviour
...
-            s_OnAuthenticationRequestFinished = new AuthorizationRequestCallback(AuthorizationRequestReceived);
-            _SetAuthorizationRequestReceivedDelegate(s_OnAuthenticationRequestFinished);
+    #if DEVELOPMENT_BUILD
+            {
+                var nativeSize = _NativeSizeof_iOSNotificationAuthorizationData();
+                var managedSize = Marshal.SizeOf(typeof(iOSAuthorizationRequestData));
+                if (nativeSize != managedSize)
+                {
+                    var error = string.Format("Native/managed struct size missmatch: {0} vs {1}", nativeSize, managedSize);
+                    Debug.LogError(error);
```
todi1856, line 96: looks like overkill, logerror and throw exception.

aurimasc: Only throw exception?
This is a critical bug, if size doesn't match.

todi1856: yes, only throw exception, since it will print the issue to log already, so no point in printing the same with Debug.LogException

aurimasc: Fixed.

com.unity.mobile.notifications/Runtime/iOS/iOSNotificationsWrapper.cs
aurimasc: Working on it right away. But the thing is quite dreaded, it was a good idea to do it in separate PR.

todi1856: Thanks, i'll approve this PR then

todi1856 (approved):

vaidasma (approved):

aurimasc:
@lucianUnity3D

PR #78, opened 2020-11-27
Notification data marshalling

aurimasc:
A set of improvement to struct marshalling between native and managed:
- Avoid dynamic struct allocation when possible
- Dynamically allocate arrays and pass to managed side atomically, when multiple structs need marshalling
- Some code cleanup
- Safety checks for all structs

### DIFF
--- a/com.unity.mobile.notifications/Runtime/iOS/Plugins/UnityNotificationManager.m
+++ b/com.unity.mobile.notifications/Runtime/iOS/Plugins/UnityNotificationManager.m
@@ -147,11 +147,13 @@
 - (void)userNotificationCenter:(UNUserNotificationCenter *)center willPresentNotification:(UNNotification *)notification
     withCompletionHandler:(void (^)(UNNotificationPresentationOptions options))completionHandler
 {
-    iOSNotificationData* notificationData = NULL;
+    iOSNotificationData notificationData;
+    BOOL haveNotificationData = NO;
     if (self.onNotificationReceivedCallback != NULL)
     {
         notificationData = UNNotificationRequestToiOSNotificationData(notification.request);
-        self.onNotificationReceivedCallback(*notificationData);
+        haveNotificationData = YES;
+        self.onNotificationReceivedCallback(notificationData);
     }
 
     BOOL showInForeground;
@@ -161,11 +163,14 @@
     {
         if (self.onRemoteNotificationReceivedCallback != NULL)
         {
-            if (notificationData == NULL)
+            if (!haveNotificationData)
+            {
                 notificationData = UNNotificationRequestToiOSNotificationData(notification.request);
+                haveNotificationData = YES;
+            }
 
             showInForeground = NO;
-            self.onRemoteNotificationReceivedCallback(*notificationData);
+            self.onRemoteNotificationReceivedCallback(notificationData);
         }
         else
         {
@@ -179,8 +184,8 @@
         showInForeground = [[notification.request.content.userInfo objectForKey: @"showInForeground"] boolValue];
     }
 
-    freeiOSNotificationData(notificationData);
-    notificationData = NULL;
+    if (haveNotificationData)
+        freeiOSNotificationData(&notificationData);
 
     if (showInForeground)
         completionHandler(presentationOptions);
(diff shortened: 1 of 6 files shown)


vaidas-unity (approved):

com.unity.mobile.notifications/Runtime/iOS/Plugins/UnityNotificationManager.m
```diff
@@ -179,8 +184,8 @@ - (void)userNotificationCenter:(UNUserNotificationCenter *)center willPresentNot
         showInForeground = [[notification.request.content.userInfo objectForKey: @"showInForeground"] boolValue];
     }
 
-    freeiOSNotificationData(notificationData);
-    notificationData = NULL;
+    if (haveNotificationData)
+        freeiOSNotificationData(&notificationData);
```
vincent-unity, line 188: Why do we have to free the memory on the stack?

vincent-unity: Oh, you want to release the heap memory held by `notificationData`, can we add some comment here?

com.unity.mobile.notifications/Runtime/iOS/iOSNotificationsWrapper.cs
```diff
@@ -169,63 +156,53 @@ public static iOSNotificationSettings GetNotificationSettings()
...
 
         public static iOSNotificationData[] GetScheduledNotificationData()
         {
 #if UNITY_IOS && !UNITY_EDITOR
-            var size = _GetScheduledNotificationDataCount();
+            int count;
+            var ptr = _GetScheduledNotificationDataArray(out count);
+            return MarshalAndFreeNotificationDataArray(ptr, count);
+#else
+            return null;
```
vincent-unity, line 181: Here we return a null instead of an empty array if scheduled notification data is empty. Not sure how this would impact the user code, as previously users can only check if the length is 0, but now they have to check if it's null before checking the length...

Same to GetDeliveredNotificationData.

aurimasc: These are internal methods. I've handled null cases where they are called.

com.unity.mobile.notifications/Runtime/iOS/Plugins/UnityNotificationManager.m
todi1856: Why we have a separate function freeiOSNotificationData, why not simply have a function inside iOSNotificationData ?

Because this code looks misleading, it looks like it's freeing the whole notificationData while in reality it frees the memory for fields.

Can you move freeiOSNotificationData to iOSNotificationData instead? That would be more clearer

todi1856 (changes requested):
Do we have any tests which cover these changes?

com.unity.mobile.notifications/Runtime/iOS/Plugins/UnityNotificationManager.m
aurimasc: That would require substantial changes in the package. This and most others are Object-C files, not Objective-C++, we cannot have methods in structs.

todi1856: oh didn't know that... do you know of any reason why all these files are objc and not objc++?

todi1856 (approved):

com.unity.mobile.notifications/Runtime/iOS/Plugins/UnityNotificationManager.m
aurimasc: I can only guess, that it was convenient at the time: XCode let's you add new Obj-C file, but not Obj-C++, for the later you have to manually rename the file extension. And you don't need to write extern "C" clutter for functions called from C# in Obj-C. The price is that you can't use C++ features, but there's almost no real need for any.

vincent-unity (approved):

vaidas-unity (approved):

vaidasma (approved):

PR #80, opened 2021-02-24
case 1315365 fix getter for time interval

aurimasc:
Fix case 1315365 - iOS Time inverval trigger getter is inconsistent with setter.
Add a couple of tests.

### DIFF
--- a/com.unity.mobile.notifications/Runtime/iOS/iOSNotificationTriggers.cs
+++ b/com.unity.mobile.notifications/Runtime/iOS/iOSNotificationTriggers.cs
@@ -90,7 +90,7 @@ namespace Unity.Notifications.iOS
         internal int timeInterval;
 
         /// <summary>
-        /// Time interval after which the notification should be delivered.
+        /// Time interval after which the notification should be delivered (only total of full seconds is considered).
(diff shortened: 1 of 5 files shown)


com.unity.mobile.notifications/Runtime/iOS/iOSNotificationTriggers.cs
```diff
@@ -94,7 +94,7 @@ public struct iOSNotificationTimeIntervalTrigger : iOSNotificationTrigger
         /// </summary>
         public TimeSpan TimeInterval
```
todi1856, line 95: Can you update inline documentation or improve property name, since right now nothing indicates to the user of TimeInterval, that it's in seconds.

todi1856 (changes requested):

com.unity.mobile.notifications/Runtime/iOS/iOSNotificationTriggers.cs
aurimasc: Since it's a public API, I think it's too minor to change anything more than documentation.

todi1856 (approved):

PR #81, opened 2021-02-25
Case 1288346: iOS Notification Settings don't get printed in the test project

aurimasc:
The reason is stripping. Because properties are not used, they get stripped and then reflection does not find them.
Unfortunately link.xml is not supported in packages.
Introduced field printing support instead.

### DIFF
--- a/TestProjects/com.unity.mobile-notifications-sample/Scripts/Helpers/Logger.cs
+++ b/TestProjects/com.unity.mobile-notifications-sample/Scripts/Helpers/Logger.cs
@@ -90,5 +90,18 @@ namespace Unity.Notifications.Tests.Sample
 
             return this;
         }
+
+        public Logger Fields(object obj, int tabs = 0)
+        {
+            foreach (var field in obj.GetType().GetFields(BindingFlags.Instance|BindingFlags.NonPublic|BindingFlags.Public))
+            {
+                var value = field.GetValue(obj).ToString();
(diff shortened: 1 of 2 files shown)


TestProjects/com.unity.mobile-notifications-sample/Scripts/Helpers/Logger.cs
```diff
@@ -90,5 +90,18 @@ public Logger Properties(object obj, int tabs = 0)
...
         }
+
+        public Logger Fields(object obj, int tabs = 0)
+        {
+            foreach (var field in obj.GetType().GetFields(BindingFlags.Instance | BindingFlags.NonPublic | BindingFlags.Public))
+            {
+                var value = field.GetValue(obj).ToString();
+                if (string.IsNullOrEmpty(value))
+                    continue;
+                Blue($"{field.Name}: {value}");
```
todi1856, line 101: don't you need to pass tabs to Blue function  ? :)

todi1856 (changes requested):

aurimasc:
> Current fix is working, however, I'm assuming you might want to take into consideration Tomas' adjustment?

Haven't I fixed that with the last commit?

vaidasma:
Oh okay, I thought maybe you haven't seen it yet

vaidasma (approved):

todi1856 (approved):

PR #82, opened 2021-03-02
Remove hardcoded notification identifier for location notifications

itsneil:
We were using "Headquarters" for every single location notification which meant that every single notification was triggered when you entered the location for just one of them. We now use the Identifier.

### DIFF
--- a/com.unity.mobile.notifications/Runtime/iOS/Plugins/UnityNotificationManager.m
+++ b/com.unity.mobile.notifications/Runtime/iOS/Plugins/UnityNotificationManager.m
@@ -290,6 +290,7 @@ bool validateAuthorizationStatus(UnityNotificationManager* manager)
     // TODO add a way to specify custom sounds.
     content.sound = [UNNotificationSound defaultSound];
 
+    NSString* identifier = [NSString stringWithUTF8String: data->identifier];
     // Generate UNNotificationTrigger from iOSNotificationData.
     UNNotificationTrigger* trigger;
     if (data->triggerType == TIME_TRIGGER)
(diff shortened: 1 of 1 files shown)


com.unity.mobile.notifications/Runtime/iOS/Plugins/UnityNotificationManager.m
```diff
@@ -320,7 +321,7 @@ - (void)scheduleLocalNotification:(iOSNotificationData*)data
         CLLocationCoordinate2D center = CLLocationCoordinate2DMake(data->locationTriggerCenterX, data->locationTriggerCenterY);
 
         CLCircularRegion* region = [[CLCircularRegion alloc] initWithCenter: center
-                                    radius: data->locationTriggerRadius identifier: @"Headquarters"];
+                                    radius: data->locationTriggerRadius identifier: identifier];
```
aurimasc, line 324: Why not create the NSString here? Would avoid allocation for other trigger types.

aurimasc (changes requested):

com.unity.mobile.notifications/Runtime/iOS/Plugins/UnityNotificationManager.m
itsneil: It was already created and being used on line 339 and is required for other trigger types, I moved the creation up the class so we could access it here.

com.unity.mobile.notifications/Runtime/iOS/Plugins/UnityNotificationManager.m
```diff
@@ -334,7 +335,6 @@ - (void)scheduleLocalNotification:(iOSNotificationData*)data
         return;
     }
 
-    NSString* identifier = [NSString stringWithUTF8String: data->identifier];
     UNNotificationRequest* request = [UNNotificationRequest requestWithIdentifier: identifier content: content trigger: trigger];
```
itsneil, line 338: @aurimasc we need to use identifier for everything else here

aurimasc (approved):

PR #83, opened 2021-03-17
case 1318503 Make iOS headers compatible only with iOS

aurimasc:
Change plugin settings for iOS header files to make them iOS-only. Before they were any platform, so they got copied into builds of all platforms, causing problems on XBox builds (at least).


vaidasma (approved):
Fixed on Xbox

PR #85, opened 2021-04-05
Include the userInfo dictionary in the notificationData.

jfperusse-bhvr:
Sometimes, third party push notifications include important information outside of a "data" userInfo key, which is why we expose userInfo directly.


aurimasc:
This is duplicate of https://github.com/Unity-Technologies/com.unity.mobile.notifications/pull/87
And I don't think exposing string and using JSON is the best approach to implement it.

jfperusse-bhvr:
Thanks for implementing this!

PR #87, opened 2021-04-09
Expose userinfo

aurimasc:
Expose UserInfo dictionary to user (case 1263498).

### DIFF
--- a/TestProjects/com.unity.mobile-notifications-sample/Scripts/ScriptableObjects/iOSNotificationTemplateTimeTrigger.cs
+++ b/TestProjects/com.unity.mobile-notifications-sample/Scripts/ScriptableObjects/iOSNotificationTemplateTimeTrigger.cs
@@ -36,6 +36,26 @@ namespace Unity.Notifications.Tests.Sample
         [Header("Time Trigger")]
         public Int32 TimeTriggerInterval;
         public bool Repeats = false;
+
+        [Serializable]
+        public struct UserDataItem
+        {
+            public string Key;
+            public string Value;
+        }
+        public UserDataItem[] UserInfo;
+
+        public Dictionary<string, string> UserInfoDict
+        {
+            get
+            {
+                var dict = new Dictionary<string, string>();
+                if (UserInfo != null)
+                    foreach (var item in UserInfo)
+                        dict.Add(item.Key, item.Value);
+                return dict;
+            }
+        }
 #endif
     }
 }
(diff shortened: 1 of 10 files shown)


TestProjects/com.unity.mobile-notifications-sample/Scripts/ScriptableObjects/iOSNotificationTemplateTimeTrigger.cs
```diff
@@ -36,6 +36,26 @@ public class iOSNotificationTemplateTimeTrigger : ScriptableObject
...
         public Int32 TimeTriggerInterval;
         public bool Repeats = false;
+
+        [Serializable]
+        public struct UserDataItem
+        {
+            public string Key;
+            public string Value;
+        }
+        public UserDataItem[] UserInfo;
```
todi1856, line 46: This variable doesn't look like it supposed be accessed, since it looks like you're indenting for user to access UserInfoDict? 

If so, maybe
[SerializeField]
private UserDataItem[] m_UserInfo;

and rename UserInfoDict to UserInfo.

Please ignore if I am wrong

com.unity.mobile.notifications/Runtime/iOS/iOSNotification.cs
```diff
@@ -181,8 +179,13 @@ public int Badge
         /// </summary>
         public string Data
         {
-            get { return data.data; }
-            set { data.data = value; }
+            get { return userInfo["data"]; }
+            set { userInfo["data"] = value; }
+        }
+
+        public Dictionary<string, string> UserInfo
```
todi1856, line 186: Is user allowed to change this data? if not - consider replacing with IReadOnlyDictionary<string, string>

todi1856 (changes requested):

com.unity.mobile.notifications/Runtime/iOS/iOSNotification.cs
aurimasc: It is allowed to modify the dictionary. Added API doc in latest commit.

TestProjects/com.unity.mobile-notifications-sample/Scripts/ScriptableObjects/iOSNotificationTemplateTimeTrigger.cs
aurimasc: This is sample-only file. This field needs to be either public or serializable so it can be edited in inspector.

todi1856 (approved):
No tests?

vaidasma (approved):

PR #95, opened 2021-06-08
Print notification in Start

aurimasc:
Print received notification details on app launch.


vaidasma:
• Notification Not Found is thrown if you bring app to foreground after sending multiple notifications (this is no longer the case if you open the app by clicking a single notification)
• Notification Not Found is thrown if you click on grouped notifications to open the app

Is it not possible to determine whether user came from a single notification and only then inform user whether the notification is found or not?

vaidasma (approved):

PR #97, opened 2021-06-10
Update auth options

aurimasc:
Updates enum for Authorization options.
Apple options: https://developer.apple.com/documentation/usernotifications/unauthorizationoptions?language=objc

Change default option for test app to Provisional. This means the app will be able to send notifications right away, but they will not show up, nor play sounds, just silently received by the app until you request non-provisional auth.


vaidasma (approved):
Silent notifications are received in the application before granting the Notification permission

PR #98, opened 2021-06-10
Check to see if we actually need to write any changes to the Preprocssor and/or Plist

itsneil:
We currently just overwrite the Plist if we need to or not, this causes Xcode to recompile each time which is a source of pain for our users. This PR is a simple set of checks for each item so we only write when needed.


aurimasc (approved):
Approved, but please don't merge in yet.

itsneil:
@aurimasc got a little extra commit here to check

aurimasc (approved):

vaidasma (approved):

PR #100, opened 2021-07-09
Fix crash in iOS push notification (case 1348518)

aurimasc:


com.unity.mobile.notifications/Runtime/iOS/Plugins/UnityNotificationData.m
```diff
@@ -213,7 +209,7 @@ void freeiOSNotificationData(iOSNotificationData* notificationData)
 
 void _ReadNSDictionary(void* csDict, void* nsDict, void (*callback)(void* csDcit, const char*, const char*))
 {
-    NSDictionary* dict = (__bridge NSDictionary*)nsDict;
+    NSDictionary* dict = (__bridge_transfer NSDictionary*)nsDict;
```
vincent-unity, line 212: I'm trying to understand the problem, and not quite sure why freeing in _ReadNSDictionary solved the problem. Could you please explain a little bit more?

com.unity.mobile.notifications/Runtime/iOS/Plugins/UnityNotificationManager.m
```diff
@@ -168,9 +171,12 @@ - (void)userNotificationCenter:(UNUserNotificationCenter *)center willPresentNot
                 notificationData = UNNotificationRequestToiOSNotificationData(notification.request);
                 haveNotificationData = YES;
             }
+            else if (notificationData.userInfo == NULL)
+                notificationData.userInfo = (__bridge_retained void*)notificationUserInfo;
```
vincent-unity, line 175: I didn't remember the details. Not quite sure if there is a case that notificationData.userInfo is NULL but notificationUserInfo is not now. IIRC, they're both converted from `notification.request`?

com.unity.mobile.notifications/Runtime/iOS/Plugins/UnityNotificationManager.m
```diff
@@ -185,7 +191,11 @@ - (void)userNotificationCenter:(UNUserNotificationCenter *)center willPresentNot
     }
 
     if (haveNotificationData)
+    {
+        if (notificationData.userInfo != NULL) // no callback, release user info
```
vincent-unity, line 195: Not sure what `no callback` means, looking through the code, only there is a callback registered, `haveNotificationData` can be YES.

com.unity.mobile.notifications/Runtime/iOS/Plugins/UnityNotificationData.m
aurimasc: The issue was that entire notification data struct was freed too early, before the dictionary is read, resulting in crash when reading the dictionary afterwards (in some cases, since somtimes the dictionary is kept alive by objc side).
It's a bit messy and I intend to do a bigger refactoring later to untangle this thing. Just that for now the priority is to ship a fix to users, since push notifications have a major regression now.

com.unity.mobile.notifications/Runtime/iOS/Plugins/UnityNotificationManager.m
aurimasc: It may get nulled above in line 159. Calling a callback now frees the dictionary, so here we restore the dictionary for the second callback.

com.unity.mobile.notifications/Runtime/iOS/Plugins/UnityNotificationManager.m
aurimasc: userInfo is set to null above in two places after callbacks are invoked. If it's not null at this point, in means there were no callbacks registered and we need to free it manually.

com.unity.mobile.notifications/Runtime/iOS/Plugins/UnityNotificationData.m
vincent-unity: Agreed it's a bit messy. I originally thought that the lastReceivedNotification(which is UNNotification) on UnityNotificationManager would hold the reference to the dictionary, which could keep it away from being released.

Anyway, improving the lifetime control would be perfect as you mentioned.

com.unity.mobile.notifications/Runtime/iOS/Plugins/UnityNotificationManager.m
vincent-unity: But if it's no callbacks, haveNotificationData will be NO, you won't get into this code branch?

vincent-unity:
@aurimasc Have you finished the refactor? Could you explain a bit more about how the issue has been fixed?

aurimasc:
> @aurimasc Have you finished the refactor? Could you explain a bit more about how the issue has been fixed?

Yes, refactoring finished. I introduced an intermediate struct that holds data and user info dictionary and convert data from native side to it at once. So now the freeing of struct and user info can happen at the same time (where they originally were).

vincent-unity:
> > @aurimasc Have you finished the refactor? Could you explain a bit more about how the issue has been fixed?
> 
> Yes, refactoring finished. I introduced an intermediate struct that holds data and user info dictionary and convert data from native side to it at once. So now the freeing of struct and user info can happen at the same time (where they originally were).

Will this make iOSNotificationCenter.OnNotificationReceived and iOSNotificationCenter.OnRemoteNotificationReceived receive different values of userInfo dictionary? The first one has the valid userInfo, the second one doesn't?

aurimasc:
> > > @aurimasc Have you finished the refactor? Could you explain a bit more about how the issue has been fixed?
> > 
> > 
> > Yes, refactoring finished. I introduced an intermediate struct that holds data and user info dictionary and convert data from native side to it at once. So now the freeing of struct and user info can happen at the same time (where they originally were).
> 
> Will this make iOSNotificationCenter.OnNotificationReceived and iOSNotificationCenter.OnRemoteNotificationReceived receive different values of userInfo dictionary? The first one has the valid userInfo, the second one doesn't?

No, the userInfo now will be the same and valid one for both of them, so it will be parsed twice, but the content will be the same.

vincent-unity (approved):

vincent-unity:
We'd better add some comments to the intermediate struct to describe why we need it :)

vaidasma (approved):
Verified as fixed:
• Original issue (case 1348518) no longer reproduces
• App is not crashing when opening one of many spammed notifications
- App is not crashing after opening notification after unlocking the screen
- App is not crashing when app is killed prior to clicking the notification
- App is not crashing when opening the notification while the app is in foreground

vaidasma:
SImulator crashes if app is not receiving any notification data upon focusing the app, when  iOSNotificationCenter.GetLastRespondedNotification(); is executed.  This is also the case in pre-fix version

vaidasma:
App crashing without no notification data has been fixed. This is good to go.

PR #101, opened 2021-07-13
Case 1340464 - remove GC allocation every frame on Android

aurimasc:
Improve Android implementation to be more performance and not make GC allocation every frame.
Changes:
- Replace Queue with List (that alone fixes allocation when AddRange is used)
- Rotate the two lists
- Quick return if no queued notifications
- Some refactoring

To QA: try to break it (flood notifications).

### DIFF
--- a/com.unity.mobile.notifications/Runtime/Android/AndroidReceivedNotificationMainThreadDispatcher.cs
+++ b/com.unity.mobile.notifications/Runtime/Android/AndroidReceivedNotificationMainThreadDispatcher.cs
@@ -10,15 +10,15 @@ namespace Unity.Notifications.Android
     {
         private static AndroidReceivedNotificationMainThreadDispatcher instance = null;
 
-        private static List<AndroidJavaObject> s_ReceivedNotificationQueue = new List<AndroidJavaObject>();
+        private List<AndroidJavaObject> m_ReceivedNotificationQueue = new List<AndroidJavaObject>();
 
-        private static List<AndroidJavaObject> s_ReceivedNotificationList = new List<AndroidJavaObject>();
+        private List<AndroidJavaObject> m_ReceivedNotificationList = new List<AndroidJavaObject>();
 
-        internal static void EnqueueReceivedNotification(AndroidJavaObject intent)
+        internal void EnqueueReceivedNotification(AndroidJavaObject intent)
         {
-            lock (instance)
+            lock (this)
             {
-                s_ReceivedNotificationQueue.Add(intent);
+                m_ReceivedNotificationQueue.Add(intent);
             }
         }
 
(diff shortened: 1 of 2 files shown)


vaidasma (approved):
Nothing out of the ordinary noticed when using the app for an extensive amount of time while monitoring the memory consumption and the behavior of the application.

com.unity.mobile.notifications/Runtime/Android/AndroidReceivedNotificationMainThreadDispatcher.cs
```diff
@@ -34,18 +34,21 @@ public void Update()
         {
             // Note: Don't call callbacks while locking receivedNotificationQueue, otherwise there's a risk
             //       that callback might introduce an operations which would create a deadlock
-            lock (s_ReceivedNotificationQueue)
+            lock (this)
             {
-                s_ReceivedNotificationList.AddRange(s_ReceivedNotificationQueue);
-                s_ReceivedNotificationQueue.Clear();
+                if (m_ReceivedNotificationQueue.Count == 0)
```
todi1856, line 39: This feels fragile, for this code to work correctly, m_ReceivedNotificationList has to be empty, when the switch occurs. If it's not empty then out of order processing of notifications is possible. Only by looking at the whole file I understand that m_ReceivedNotificationList is empty when this switch occurs, but still if in the future someone will mess with m_ReceivedNotificationList, it will be easy to miss this place

Consider adding a safeguard
if  (m_ReceivedNotificationList.Count != 0)
    throw new Exception(..)

todi1856 (commented):
Waiting for reply

com.unity.mobile.notifications/Runtime/Android/AndroidReceivedNotificationMainThreadDispatcher.cs
aurimasc: If it's not empty, it means there are leftover notifications from last processing which shouldn't happen. Actually, this code still has a bug where this could happen (in case of exception), but that's addressed in different PR. When such leftovers happen, you get callbacks called again for the same notification next frame, most likely leading to the same exception again and postponing the same issue to the next frame etc. This change would scramble the ordering, but you'd still get bombarded with callbacks every frame when this happens.
Throwing exception like you propose would prevent callbacks from being called with logging about a bug. IMO, not throwing exception gives a clearer indication of bug (but it doesn't log the exact issue).

todi1856: "clearer indication of bug" - really? :) it's really not that apparent, especially for a green user. Well, if you're a fan of hidden bugs, so be it :)

todi1856 (approved):

com.unity.mobile.notifications/Runtime/Android/AndroidReceivedNotificationMainThreadDispatcher.cs
aurimasc: It's for sure clearer in our test project, reducing chances of us releasing a bug :)

Issue #99, opened 2021-06-21
Implicit PendingIntent Vulnerability


AntonPetrov83:
Hi,

Looks like after we have integrated Mobile Notifications 1.3.2 we received an e-mail from the Google:

> Your app(s) contain an Implicit PendingIntent issue.
> To address this issue, follow the steps in this [Google Help Center](https://notifications.google.com/g/p/AD-FnEyMqkufpTSko58ywH25jwojLR4HL9ZsYGoo48dnW3pFPbPFb-C6lsQQB2QCyYKRtBJHLKhTWxbMXcvpigRxYD_86lPSB2-zraPS2UAbTC27Eb0) article.

I see that you do not use `FLAG_IMMUTABLE` in the package. Can it be the problem?

Thank you.

chinpok:
Have a same issue but disabling mobile notification do not fix issue

chinpok:
https://forum.unity.com/threads/vulnerability-implicit-pendingintent.1131070/

AntonPetrov83:
Yeah, its devtodev sdk. They released an update.

aurimasc:
The fix for this has landed to master.

yantian001:
This issue still happened on Mobile Notifications 2.0.0 with Unity 2019.4.38f1c1
![image](https://user-images.githubusercontent.com/8159444/164884523-03435bd1-21f6-4f9c-9cde-b2262bc822b1.png)

PR #103, opened 2021-07-15
Use FLAG_IMMUTABLE for PendingIntent

aurimasc:
Ensure we don't have PendingIntent vulnerability:
https://github.com/Unity-Technologies/com.unity.mobile.notifications/issues/99


todi1856 (approved):
Would have been probably easier to create GetPendingIntentFlags function instead of creating two new functions.

Something's like:
```
int GetPendingIntentFlags (int flags)
{
   if (Build.VERSION.SDK_INT < 23)
       return flags;
   return flags | PendingIntent.FLAG_IMMUTABLE
}
```

PR #111, opened 2021-08-19
Run iOS tests on CI

aurimasc:
- Add CI for running tests on iOS (fixes the tests too)
- Breaking change in settings: the type of DefaultAuthorizationOptions changed to AuthorizationOption (probably this had to be so from beginning and was a mistake).

### DIFF
--- a/.yamato/upm-ci.yml
+++ b/.yamato/upm-ci.yml
@@ -24,6 +24,10 @@ mobile_dependencies:
     name: com.unity.mobile.notifications
     version: file:../../com.unity.mobile.notifications
 
+utr:
+  url_win: https://artifactory.prd.it.unity3d.com/artifactory/unity-tools-local/utr-standalone/utr.bat
+  url_unix: https://artifactory.prd.it.unity3d.com/artifactory/unity-tools-local/utr-standalone/utr
+
 ---
 pack:
   name: Pack
--- a/.yamato/upm-ci.yml
+++ b/.yamato/upm-ci.yml
@@ -79,7 +79,7 @@ test_iOS_{{ editor.version }}:
     image: mobile/macos-10.13-testing:stable
     flavor: b1.medium
   variables:
-    IOS_SIMULATOR_SDK=1
+    IOS_SIMULATOR_SDK: 1
   commands:
     - brew tap --force-auto-update unity/unity git@github.cds.internal.unity3d.com:unity/homebrew-unity.git
     - brew install unity-config
--- a/.yamato/upm-ci.yml
+++ b/.yamato/upm-ci.yml
@@ -78,8 +78,6 @@ test_iOS_{{ editor.version }}:
     type: Unity::mobile::iPhone
     image: mobile/macos-10.13-testing:stable
     flavor: b1.medium
-  variables:
-    IOS_SIMULATOR_SDK: 1
   commands:
     - brew tap --force-auto-update unity/unity git@github.cds.internal.unity3d.com:unity/homebrew-unity.git
     - brew install unity-config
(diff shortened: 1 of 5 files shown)


.yamato/upm-ci.yml
```diff
@@ -53,6 +79,35 @@ test_{{ platform.name }}_{{ editor.version }}:
...
+  agent:
+    type: Unity::mobile::iPhone
+    image: mobile/macos-10.13-testing:stable
+    flavor: b1.medium
+  commands:
+    - brew tap --force-auto-update unity/unity git@github.cds.internal.unity3d.com:unity/homebrew-unity.git
+    - brew install unity-config
+    - unity-config settings editor-path .Editor
+    - unity-config project create .MyTestProject
+{%- for dependency in mobile_dependencies -%}
```
todi1856, line 94: You're creating project from scratch, wouldn't it be easier for you to have a commited Unity project?

com.unity.mobile.notifications/Editor/NotificationSettings.cs
```diff
@@ -117,15 +117,15 @@ public static bool RequestAuthorizationOnAppLaunch
             /// <summary>
             /// Configure the notification interaction types your app will include in the authorisation request if RequestAuthorizationOnAppLaunch is enabled. Alternatively you can specify them when creating a `AuthorizationRequest` from a script.
             /// </summary>
-            public static PresentationOption DefaultAuthorizationOptions
+            public static AuthorizationOption DefaultAuthorizationOptions
```
todi1856, line 120: Isn't this a a breaking change? Shouldn't you have backwards compatible API present with Obsolete attribute?

com.unity.mobile.notifications/Tests/Runtime/iOS/iOSNotificationTests.cs
```diff
@@ -34,10 +75,14 @@ public void AfterEachTest()
         receivedNotificationCount = 0;
         lastReceivedNotification = null;
     }
+#endif
 
     [UnityTest]
     public IEnumerator SendSimpleNotification_NotificationIsReceived()
     {
+#if UNITY_EDITOR
```
todi1856, line 83: There's a better a way, please use UnityPlatform attribute and specify the platforms where this test should run

todi1856: You can actually specify UnityPlatform  on top of class, if this whole class is for iOS tests only

aurimasc: Good one, didn't knew about that. Fixed.

com.unity.mobile.notifications/Editor/NotificationSettings.cs
aurimasc: Unfortunately it is. I think it was originally a typo no one noticed, since the two enums have very similar names and back then they even had the same values.
My reasoning for breaking was:
- No one complained about this so probably no one used it.
- Any serialized value is serialized as int, so deserializing will result in correct values.
- This is Editor only API for manipulating settings, for which UI exists too. So someone uses it, it's probably like one place in the code and will be easy to fix.
- Obsoleting/deprecating would mean introducing new API and inventing a name for it, since the most natural name is already used.
So considering all the points above, I decided that this API breakage is the best course of action.

todi1856 (approved):

.yamato/upm-ci.yml
aurimasc: The committed project would have to be maintained for the oldest Unity version supported by package and any later versions become prone to any issues that may arise when upgrading project, while this creates a clean Unity project for given version.

PR #114, opened 2021-08-26
ShowInForeground feature for android

serefbulbul:
Hi!
Actually local notifications are received even though app is on the foreground for Android. However, there can be cases that behaviour wanted to be aligned with iOS. This was the case for our game teams. Therefore, we patched your code to have a "ShowInForeground" flag for local notifications as well and I wanted to share the solution with you here, if you are interested. 
I am also open for any kind of improvements.


com.unity.mobile.notifications/Runtime/Android/AndroidNotification.cs
```diff
@@ -177,6 +177,11 @@ public DateTime CustomTimestamp
             }
         }
 
+        /// <summary>
+        /// Set this notification to be shown when app is in the foreground.
+        /// </summary>
+        public bool ShowInForeground { get; set; }
```
aurimasc, line 183: Unfortunately this is a breaking change.
Mu suggestion is to back the ShowInForeground property by m_HideInForeground boolean, so that if you don't set the property, the default behavior is as it used to be.

com.unity.mobile.notifications/Runtime/Android/Plugins/com/unity/androidnotifications/UnityNotificationManager.java
```diff
@@ -539,7 +542,10 @@ protected static void sendNotification(Context context, Intent intent) {
 
     // Call the system notification service to notify the notification.
     protected static void notify(Context context, int id, Notification notification) {
-        getNotificationManager(context).notify(id, notification);
+        boolean showInForeground = notification.extras.getBoolean("showInForeground", false);
+        if (!isInForeground() || showInForeground) {
```
aurimasc, line 546: This isn't entirely consistent with iOS behavior. On iOS notification is not shown, but the callback is still called informing about the arrival of notification.

aurimasc (changes requested):
Looks interesting and useful.
But maintaining backward compatibility is important. And if we aim for feature parity with iOS, behavior should match as much as possible.

PR #115, opened 2021-09-08
case 1360115 fix focus loss when switching settings

aurimasc:
https://jira.unity3d.com/browse/MPT-1619

There things here:
- Fix case 1360115 - lost focus when trying to switch from notifications to different settings (like Player)
- Re-fix the bug that originally introduced this issue: https://github.com/Unity-Technologies/com.unity.mobile.notifications/pull/70
- Call base implementation in OnDeactivate, since that one is not empty

QA: check whether settings work fine (and get saved properly) after this, especially with play mode, which is being re-fixed here.

### DIFF
--- a/com.unity.mobile.notifications/Editor/NotificationSettingsProvider.cs
+++ b/com.unity.mobile.notifications/Editor/NotificationSettingsProvider.cs
@@ -5,6 +5,7 @@ using UnityEditor;
 using UnityEditorInternal;
 using Unity.Notifications.iOS;
 using UnityEngine.Assertions;
+using UnityEngine.UIElements;
 
 namespace Unity.Notifications
 {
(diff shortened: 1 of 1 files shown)


todi1856 (approved):

vaidasma:
Minor issue:
After entering playmode, the mobile notification returns to the tab which was last edited (Android/iOS), not the one which was last opened. Example scenario:
 1. Change some settings in Android Tab
 2. Switch to iOS tab
 3.  Enter Playmode
 
 Result: the tab will be switched to Android, as editor did not focus on the iOS tab (that's my guess)
 Same scenario happens if you exit Playmode while having selected the tab that's not the one that was edited last
 
 Other that that it looks good, I'll recheck it after this has been fixed to make sure there's nothing else hiding behind this issue

aurimasc:
> After entering playmode, the mobile notification returns to the tab which was last edited (Android/iOS)

Fixed in latest commit.

vaidasma (approved):

PR #116, opened 2021-09-13
Add Android CI

aurimasc:
Add CI to run tests on Android. Currently trunk and 2021.2 are excluded, since running playmode test in these is broken for Android.
Couple not related small changes in this PR:
- Refactored Android tests to use attributes instead of defines
- Remove dependency on current culture when generating notification ID (breaks on Arabic).

### DIFF
--- a/.yamato/upm-ci.yml
+++ b/.yamato/upm-ci.yml
@@ -177,6 +177,9 @@ test_trigger:
     {% for editor in ios_test_editors %}
     - .yamato/upm-ci.yml#test_iOS_{{ editor.version }}
     {% endfor %}
+    {% for editor in android_test_editors %}
+    - .yamato/upm-ci.yml#test_Android_{{ editor.version }}
+    {% endfor %}
 
 publish:
   name: Publish to Internal Registry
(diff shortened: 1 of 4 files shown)


.yamato/upm-ci.yml
```diff
@@ -79,6 +85,40 @@ test_{{ platform.name }}_{{ editor.version }}:
 {% endfor %}
 {% endfor %}
 
+{% for editor in android_test_editors %}
```
todi1856, line 88: I think usually it's recommended to Split & Run for Android devices.

Since now, while the Editor is building your stuff you're allocating Android device, which is wasteful.

See https://github.cds.internal.unity3d.com/unity/mobile-yamato-example

com.unity.mobile.notifications/Runtime/Android/AndroidNotificationCenter.cs
```diff
@@ -212,7 +212,10 @@ public static int SendNotification(AndroidNotification notification, string chan
             if (!Initialize())
                 return -1;
 
-            int id = Math.Abs(DateTime.Now.ToString("yyMMddHHmmssffffff").GetHashCode()) + (new System.Random().Next(10000));
+            // Now.ToString("yyMMddHHmmssffffff"), but avoiding any culture-related formatting or dependencies
+            var now = DateTime.UtcNow;
+            var nowFormatted = $"{now.Year}{now.Month}{now.Day}{now.Hour}{now.Minute}{now.Second}{now.Millisecond}";
+            int id = Math.Abs(nowFormatted.GetHashCode()) + (new System.Random().Next(10000));
```
todi1856, line 218: Wondering if this new System.Random().Next(10000)) won't cause a situation where id in ultra rare cases will be the same

todi1856 (approved):

PR #118, opened 2021-09-17
Add support for iOS notification attachments

aurimasc:
Add support for notification attachments on iOS. Attachments let you add images, sounds of movies to notifications.
For now leaving out attachment options: implementing support for them is much harder and it's not certain, how useful they are.

QA:
- I've only tested with image (the implementation is the same, just provide URL to file)
- Do you think it would be useful to add a button to test project specifically for this?


com.unity.mobile.notifications/Runtime/iOS/iOSNotification.cs
```diff
@@ -380,7 +390,24 @@ internal iOSNotificationWithUserInfo GetDataForSending()
...
             return ret;
         }
+
+        /// <summary>
+        /// Add an attachment to notification.
+        /// </summary>
+        /// <param name="attachment">Attachment to add</param>
+        public void AddAttachment(iOSNotificationAttachment attachment)
+        {
+            var attachments = Attachments;
```
todi1856, line 403: I don't understand why we need AddAttachment  function?

Maybe initialize Attachments in in iOSNotification constructor, then as I user I can simply do

var notification = new iOSNotification ();
notifacation.Attachements.Add(...)

com.unity.mobile.notifications/Runtime/iOS/iOSNotificationAttachment.cs
```diff
@@ -0,0 +1,20 @@
+namespace Unity.Notifications.iOS
+{
+    /// <summary>
+    /// Notification attachment.
+    /// Refer to Apple documentation for details.
+    /// </summary>
+    /// <see cref="https://developer.apple.com/documentation/usernotifications/unnotificationattachment?language=objc"/>
+    public struct iOSNotificationAttachment
```
todi1856, line 8: If it's not allowed to change Id/Url after the creation of iOSNotificationAttachment, consider doing this:

```
  public struct iOSNotificationAttachment
  {
        /// <summary>
        /// A unique identifier for the attachments. Will be auto-generated if left empty.
        /// </summary>
        public string Id { get; }

        /// <summary>
        /// URL to local file, accessible to the application.
        /// </summary>
        public string Url { get; }
       public iOSNotificationAttachment(string id, string url)
       {
           Id = id;
           Url = url;
       }
    }
```

com.unity.mobile.notifications/Runtime/iOS/iOSNotificationsWrapper.cs
```diff
@@ -301,6 +328,32 @@ public static IntPtr CsDictionaryToObjC(Dictionary<string, string> userInfo)
 #endif
         }
 
+        public static IntPtr CsAttachmentsToObjc(List<iOSNotificationAttachment> attachments)
+        {
+#if UNITY_IOS && !UNITY_EDITOR
+            if (attachments == null)
```
todi1856, line 334: could probably add additional check here attachments == null || attachments.Count == 0

todi1856 (changes requested):

com.unity.mobile.notifications/Runtime/iOS/iOSNotification.cs
aurimasc: I don't expect attachments to be frequently used, so I'd like to avoid the allocation. So I added this for convenience.

com.unity.mobile.notifications/Runtime/iOS/iOSNotificationAttachment.cs
aurimasc: It's fine to change them.

com.unity.mobile.notifications/Runtime/iOS/iOSNotificationsWrapper.cs
aurimasc: Good idea.

com.unity.mobile.notifications/Runtime/iOS/iOSNotification.cs
todi1856: Even so, this looks wrong, since you're providing with two ways on how to add attachements, if you're worried about allocations, why not do this then:
* Remove AddAttachment
* Change public List<iOSNotificationAttachment> Attachments { get; set; } to
```
List<iOSNotificationAttachment> m_Attachements;
public List<iOSNotificationAttachment> Attachments {
{
get 
{
   if (m_Attachements == null)
       m_Attachements = new List<iOSNotificationAttachment>();
   return m_Attachements ; 
}
set
{
   m_Attachements  = value;
}

}
```

todi1856 (changes requested):

com.unity.mobile.notifications/Runtime/iOS/iOSNotification.cs
aurimasc: But this make querying attachments create a list.
From my expected usage AddAttachment would be enough, but I don't want to limit users to not being able remove that added attachments.

todi1856: I am not sure I understand why you're trying so hard here to avoid allocation, since as far as I understand, attaching attachements is a very rare operation, on other hand, you're creating this inconsistent API, where you have AddAttachement function and yet you also have a property where you can also manipulate the list of attachements..

Here's another suggestion you could do:
* Add RemoveAttachement
* Change public List<iOSNotificationAttachment> Attachments to IReadOnlyList<iOSNotifcationAttachemen>

That way there wouldn't duplication of operations on API side.


But in reality, I would simply leave List<iOSNotificationAttachment> Attachments property and initialize it in iOSNotification,  since user needs to instantiate iOSNotification anyways, right? This would avoid this all unecessary complexity in your code

todi1856 (approved):

vaidasma (approved):

PR #120, opened 2021-09-24
Reinitialize manager if gone (domain reload)

aurimasc:
If you build player with Notification Settings open, after domain reload m_SettingsManager will be null and needs to be reinitialized.


todi1856 (approved):
weird that you need this, I haven't encountered this problem with logcat package

PR #124, opened 2021-10-14
Add APIs to open notification settings

aurimasc:
https://jira.unity3d.com/browse/MPT-626

Adds APIs to open notification settings.
On iOS there is no way to open notification settings specifically, so we open app settings (as close enough).
Likewise on pre-Oreo (8.0) Androids.
On Android 8 and later the API opens notification setting for the app or for a specific channel, if not null.


vaidasma:
Feature working as intended. App settings / Notification settings are opened correctly on all tested devices:
• Xiaomi Mi 5s (OS:6.0.1)
• Samsung Galaxy S8 (OS:7)
• OnePlus OnePlus 6T OS:9)
• Google Pixel 5 (OS:12)

• iPhone 6s (OS:11)
• iPhone 12 (OS:15.2)

vaidasma (approved):

PR #126, opened 2021-10-21
Case 1373835 fix calendar notification for buddhist calendars and related improvements

aurimasc:
https://jira.unity3d.com/browse/MPT-1731

Fix calendar notifications on iOS on devices with buddhist calendar and related changes:
- Refactored notification data struct to be more compact
- Refactored triggers by making Type property instance to avoid typechecks; this is technically a breaking change, but previous static properties were completely useless to users, unlikely anyone used them (and we already had a breaking change in the package since last release, good opportunity for another minor one)
- On native side we now use UTC calendar; on managed side we still default to local, but give users option to use UTC too
- Added tests


vaidasma (approved):
Devices with buddhist calendar being used can now receive notifications. Tested on user-provided project. The project needed the Lunar Console update as well as some code changes to work 

Tested with devices:
• iPhone 6s (OS:14.3)
• iPhone 12 (OS:15)

PR #129, opened 2021-11-12
Notification drawables management

aurimasc:


vaidasma (approved):
All editor operations: adding / removing / clearing icons via available methods are working as intended.

Settings window now also properly updates after executing one of the actions mentioned above.

PR #130, opened 2021-11-12
Fixes an issue where app hangs on startup with both Unity notifications and Firebase in the project

itsneil:
https://fogbugz.unity3d.com/f/cases/1375744/

We need to move the setting of a delegate out of the load method in order to fix the above case.

### DIFF
--- a/com.unity.mobile.notifications/Runtime/iOS/Plugins/UnityAppController+Notifications.mm
+++ b/com.unity.mobile.notifications/Runtime/iOS/Plugins/UnityAppController+Notifications.mm
@@ -46,10 +46,13 @@
              [UnityNotificationManager sharedInstance].lastReceivedNotification = NULL;
          }];
 
-        [nc addObserverForName: UIApplicationDidFinishLaunchingNotification
+        [nc addObserverForName: kUnityWillFinishLaunchingWithOptions
          object: nil
          queue: [NSOperationQueue mainQueue]
          usingBlock:^(NSNotification *notification) {
+
+             [UNUserNotificationCenter currentNotificationCenter].delegate = [UnityNotificationManager sharedInstance];
+
              BOOL authorizeOnLaunch = [[[NSBundle mainBundle] objectForInfoDictionaryKey: @"UnityNotificationRequestAuthorizationOnAppLaunch"] boolValue];
              BOOL supportsPushNotification = [[[NSBundle mainBundle] objectForInfoDictionaryKey: @"UnityAddRemoteNotificationCapability"] boolValue];
              BOOL registerRemoteOnLaunch = supportsPushNotification == YES ?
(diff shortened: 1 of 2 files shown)


itsneil:
After discussions with Alexey, it was noted that setting the delegate in load is a bad bad idea.

https://developer.apple.com/documentation/objectivec/nsobject/1418815-load?language=objc
https://www.mikeash.com/pyblog/friday-qa-2009-05-22-objective-c-class-loading-and-initialization.html

Adding them to the PR

itsneil:
Sorry for the git mess, GUI client... When squashed it will be sensible.

aurimasc:
@vaidasma after the changes the concerning points are push notifications and app launch by clicking the notification. By changing the place when we set the delegate we risk losing info about notifications delivered while app was in background (or killed entirely) and not receiving information about the notification that was used to launch it.

alexey-unity:
> are push notifications and app launch by clicking the notification

if that change results in issues or we worry about resulting in the issues - should we try +initialize ? this happens when app runtime is fully setup

Pavel-Unity:
- the way the notification package do its initialization seems won't work in UaaL scenario when another app is fully running so there will be no UIApplicationDidFinishLaunchingNotification even broad casted
- according to https://developer.apple.com/documentation/usernotifications/unusernotificationcenter/1649522-delegate?language=objc Notifications that cause your app to be launched or delivered shortly after (application:didFinishLaunchingWithOptions) methods finish executing
 but we are attaching initialization code not to apps delegate callback but to message broadcasted UIApplicationDidFinishLaunchingNotification. 
 The question is it guaranteed that UIApplicationDidFinishLaunchingNotification will be broadcasted before UNUserNotificationCenter.delegate calls start to happen

itsneil:
Given https://developer.apple.com/documentation/usernotifications/unusernotificationcenter/1649522-delegate?language=objc 

> To guarantee that your app responds to all actionable notifications, you must set the value of this property before your app finishes launching. For an iOS app, this means updating this property in the application:willFinishLaunchingWithOptions: or application:didFinishLaunchingWithOptions: method of the app delegate.

And in order to allow UaaL to function correctly, I have decided to have the notifications init on the broadcast of kUnityWillFinishLaunchingWithOptions.

When launching an app, the AppDelegate's methods are called and then the notification posted shortly afterward, we cannot guarantee the timing of the native lifecycle notifications, but we always post kUnityWillFinishLaunchingWithOptions from the AppDelegate. This could guarantee that the notification stuff is set up late enough not to cause locks and early enough not to miss notifications.

This would also be a good thing to document to all plugin writers as a good place to init your plugin?

Pavel-Unity (approved):

alexey-unity (approved):

vaidasma (approved):
Verified as fixed:
• No longer crashes upon killing the app and launching again (with FIrebase used in the project)
• Application does not demonstrate any unwanted behavior when putting it to background/foreground, locking/unlocking
• Notifications Authorization is successful
• Local/push notifications are received

Tested with Notification projects merged into the user's project from Fogbugz
Device used: iPhone 12 (OS:15.2)

Note: during testing, it has been noticed that authorization on launch no longer works. Unrelated to the current fix.

aurimasc (approved):

aurimasc:
@vaidasma after the changes the concerning points are push notifications and app launch by clicking the notification. By changing the place when we set the delegate we risk losing info about notifications delivered while app was in background (or killed entirely) and not receiving information about the notification that was used to launch it.

vaidasma:
Above issue by design: caused by "Provisional" flag in Default Notification Authorization Settings

seagullua:
When the fix will become available in Unity Package Manager?

PR #134, opened 2021-11-30
[ios] case 1382960 handle numeric data in user info

aurimasc:
https://jira.unity3d.com/browse/MPT-1796
Fix crash when push notification JSON file has integer item at the top level.

No automated test yet, need to figure out how to automate push notification testing, task for it:
https://jira.unity3d.com/browse/MPT-1797


todi1856 (approved):

miv:
@aurimasc I'm sorry to barge in, but i think this is a bad fix. Apple clearly states what types can be in it's json

> Your custom keys must have values with primitive types, such as dictionary, array, string, number, or Boolean. Custom keys are available in the userInfo dictionary of the UNNotificationContent object delivered to your app.

We are using all of them in userInfo.

aurimasc:
@miv only boolean is left a bit of an issue, since now it comes as a string value "0" or "1". I'll fix it in another PR.

PR #136, opened 2021-12-01
Android12 support

aurimasc:
case 1376849
Fix using package on Android 12:
* NotificationRestartOnBootReceiver needs to be exported
* Disable exact alarms on Android 12 until we can support them

The exact alarms have been disabled because since Android 12 it requires a special permission to be requested in order to use them. Also, we need to handle it in source code, which means using APIs introduced only in API 31, which in turn requires compiling using latest SDK which isn't the one that we ship with Unity...
For now it is more important to ship new version of package that works than to fully support it. So on Android 12 we'll use inexact alarm (like we do on pre-M Androids), so notification timing will be less accurate. Users who really need exact timing will have to manually edit out source code and handle the permission stuff.


vaidasma:
Skipping QA pass, will verify in an upcoming release

todi1856 (approved):

PR #152, opened 2022-02-10
Fix publishing to google play

luvjungle:
Recent builds get rejected by Google Play with an error:
"You uploaded an APK or Android App Bundle which has an activity, activity alias, service or broadcast receiver with intent filter, but without 'android:exported' property set. This file can't be installed on Android 12 or higher. See: developer.android.com/about/versions/12/behavior-changes-12#exported"

This MR fixes this error by adding needed attribute to notificationRestartOnBootReceiver

### DIFF
--- a/com.unity.mobile.notifications/Editor/AndroidNotificationPostProcessor.cs
+++ b/com.unity.mobile.notifications/Editor/AndroidNotificationPostProcessor.cs
@@ -119,6 +119,7 @@ namespace Unity.Notifications
             {
                 notificationRestartOnBootReceiver = manifestXmlDoc.CreateElement("receiver");
                 notificationRestartOnBootReceiver.SetAttribute("name", kAndroidNamespaceURI, kNotificationRestartOnBootName);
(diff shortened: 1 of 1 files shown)


com.unity.mobile.notifications/Editor/AndroidNotificationPostProcessor.cs
```diff
@@ -119,6 +119,7 @@ internal static void InjectReceivers(string manifestPath, XmlDocument manifestXm
             {
                 notificationRestartOnBootReceiver = manifestXmlDoc.CreateElement("receiver");
                 notificationRestartOnBootReceiver.SetAttribute("name", kAndroidNamespaceURI, kNotificationRestartOnBootName);
+                notificationRestartOnBootReceiver.SetAttribute("exported", kAndroidNamespaceURI, "false");
```
aurimasc, line 122: Why is this necessary?
This same attribute should be set a bit below, on line 133, except that it probably needs to be changes to false.

aurimasc (commented):
Under what circumstances the line 133 doesn't do it's job?

PR #158, opened 2022-03-16
Min android api update

aurimasc:
The minimum Android API supported by package is 21, but that was missed when releasing 2.0.0.
Also, a couple of changes were missed to and were written to Unity release notes when verifying.

### DIFF
--- a/com.unity.mobile.notifications/Editor/AndroidNotificationPostProcessor.cs
+++ b/com.unity.mobile.notifications/Editor/AndroidNotificationPostProcessor.cs
@@ -15,9 +15,23 @@ namespace Unity.Notifications
 
         public void OnPostGenerateGradleAndroidProject(string projectPath)
         {
+            MinSdkCheck();
+
             CopyNotificationResources(projectPath);
 
             InjectAndroidManifest(projectPath);
+        }
+
+        private void MinSdkCheck()
+        {
+#if !UNITY_2020_2_OR_NEWER
+        // API level 21 not supported since 2020.2, need to check for prior releases
+        const AndroidSdkVersions kMinAndroidSdk = AndroidSdkVersions.AndroidApiLevel21;
+
+        if (PlayerSettings.Android.minSdkVersion < AndroidSdkVersions.AndroidApiLevel21)
+                throw new NotSupportedException(string.Format("Minimum Android API level supported by notifications package is {0}, your Player Settings have it set to {1}",
+                    (int)kMinAndroidSdk, PlayerSettings.Android.minSdkVersion));
+#endif
         }
 
         private void CopyNotificationResources(string projectPath)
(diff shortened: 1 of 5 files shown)


com.unity.mobile.notifications/Editor/AndroidNotificationPostProcessor.cs
```diff
@@ -15,11 +15,25 @@ public class AndroidNotificationPostProcessor : IPostGenerateGradleAndroidProjec
...
         }
 
+        private void MinSdkCheck()
+        {
+#if !UNITY_2020_2_OR_NEWER
+        // API level 21 not supported since 2020.2, need to check for prior releases
+        const AndroidSdkVersions kMinAndroidSdk = AndroidSdkVersions.AndroidApiLevel21;
+
+        if (PlayerSettings.Android.minSdkVersion < AndroidSdkVersions.AndroidApiLevel21)
+                throw new NotSupportedException(string.Format("Minimum Android API level supported by notifications package is {0}, your Player Settings have it set to {1}",
```
todi1856, line 32: maybe it's better to use https://docs.unity3d.com/ScriptReference/Build.IPreprocessBuildWithReport.OnPreprocessBuild.html, since it looks like you want to fail early, instead of making user wait for the whole build to finish and only then fail?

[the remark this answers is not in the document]
aurimasc: Is interpolation supported on all Unity releases? I remember something not being supported in 19.4.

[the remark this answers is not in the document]
todi1856: I use interpolation in logcat package, and my tests are green with 19.4. I would guess yes - it's supported. But if you have doubts, you can leave it as it is

aurimasc: Huh, if I throw exception out of that one, it logs it to console, but still continues with a build and produces apk!

todi1856: Huh, that's strange, maybe worth asking in #devs-asset-build-pipeline, sounds like a bug. I'll approve your PR, if this is blocking you

todi1856 (approved):

com.unity.mobile.notifications/Editor/AndroidNotificationPostProcessor.cs
rusitschka: Try BuildPlayerWindow.BuildMethodException

aurimasc: Same issue. Maybe it's fixed in newer Unity, but I need this specifically for 19.4

paradizIscool:
Please, don't drop support for Android 19

PR #159, opened 2022-03-16
Cleanup sample

aurimasc:
Merge the two directories together.
Originally we had sample project as package with per-version directories. With only one variant left this looks redundant.

QA: there are no package changes here, just try randomly the test project on both platforms.


vaidasma (approved):
Verified:
• Android: tested different types of notifications from the test project
• iOS: tested different types of notifications from the test project, including push notifications

Issue #160, opened 2022-03-18
Android 12 bug. version 2.0.0 not working


tiennguyenlastmile:
Fatal Exception: java.lang.RuntimeException
Unable to start receiver com.unity.androidnotifications.UnityNotificationRestartOnBootReceiver: java.lang.IllegalArgumentException:  Targeting S+ (version 31 and above) requires that one of FLAG_IMMUTABLE or FLAG_MUTABLE be specified when creating a PendingIntent. Strongly consider using FLAG_IMMUTABLE, only use FLAG_MUTABLE if some functionality depends on the PendingIntent being mutable, e.g. if it needs to be used with inline replies or bubbles

Please fix this issue.
Thanks advance

PR #161, opened 2022-03-22
Properly set intent immutable flag after reboot

aurimasc:
Duplicate reports:
https://github.com/Unity-Technologies/com.unity.mobile.notifications/issues/160
case 1410927

Android 12 requires Pending intents to have immutable flag, one place was missed: rechedule after reboot.
Tested fix on Asus ROG Phone (Android 8.1).

QA: try before and after fix. Note, that on some phones rescheduling is not reliable, for example on Asus ROG I had to start the app immediately after reboot for it to actually get a reschedule. It's expected that on Android 12 rescheduling will not work either way prior this fix (exception thrown, but retrieving logcat is difficult), but should work after the fix. Test project in this repo is suitable (has a button specifically for this).


vaidasma:
Tested on Google Pixel 5 (OS:12).

Rescheduling after restart works. However it also worked before fix on both of my Android 12 devices:
• Google Pixel 6
• Galaxy S10+

Resolving as fixed

vaidasma (approved):

PR #166, opened 2022-04-06
Add ShowInForeground boolean feature for Android notifications.

serefbulbul:
Hi,

Android notification pop-ups are shown even when app is on the foreground if it is received on a notification channel with high priority. However, there can be cases that behaviour is wanted to be aligned with iOS. This was the case for our game teams. Therefore, we patched your code to have a "ShowInForeground" flag for local notifications as well and I wanted to share the solution with you here.

- With ShowInForeground option, it is possible not to show notification pop-ups if app is on the foreground
- Related callbacks about receiving notification will still be called

Similar PR was opened before but we couldn't maintain it because of reasons. Now, I am open for any feedback to improve and merge this PR.

Bests,
Seref

### DIFF
--- a/com.unity.mobile.notifications/Runtime/Android/AndroidNotification.cs
+++ b/com.unity.mobile.notifications/Runtime/Android/AndroidNotification.cs
@@ -177,6 +177,11 @@ namespace Unity.Notifications.Android
             }
         }
 
+        /// <summary>
+        /// Set this notification to be shown when app is in the foreground.
+        /// </summary>
+        public bool ShowInForeground { get; set; }
+
         internal bool ShowCustomTimestamp { get; set; }
 
         private Color m_Color;
@@ -208,6 +213,7 @@ namespace Unity.Notifications.Android
             GroupAlertBehaviour = GroupAlertBehaviours.GroupAlertAll;
             ShowTimestamp = false;
             ShowCustomTimestamp = false;
+            ShowInForeground = true;
 
             m_RepeatInterval = (-1L).ToTimeSpan();
             m_Color = new Color(0, 0, 0, 0);
--- a/com.unity.mobile.notifications/Runtime/Android/Plugins/com/unity/androidnotifications/UnityNotificationManager.java
+++ b/com.unity.mobile.notifications/Runtime/Android/Plugins/com/unity/androidnotifications/UnityNotificationManager.java
@@ -1,6 +1,7 @@
 package com.unity.androidnotifications;
 
 import android.app.Activity;
+import android.app.ActivityManager;
 import android.app.AlarmManager;
 import android.app.Notification;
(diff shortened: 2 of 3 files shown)


com.unity.mobile.notifications/Runtime/Android/AndroidNotification.cs
```diff
@@ -208,6 +213,7 @@ public AndroidNotification(string title, string text, DateTime fireTime)
             GroupAlertBehaviour = GroupAlertBehaviours.GroupAlertAll;
             ShowTimestamp = false;
             ShowCustomTimestamp = false;
+            ShowInForeground = true;
```
aurimasc, line 216: This assignment only applies to this constructor, but struct also has a default no arg one you cannot write. So for default-constructed struct the value would be false.
I suggest instead to have a private variable m_SilentInForeground in struct with ShowInForeground returning/setting the opposite value from that. Then the property would have the correct value and backward compatibility is preserved.

com.unity.mobile.notifications/Runtime/Android/Plugins/com/unity/androidnotifications/UnityNotificationManager.java
```diff
@@ -53,6 +56,7 @@
     protected static final String KEY_NOTIFICATION = "unityNotification";
     protected static final String KEY_SMALL_ICON = "smallIcon";
     protected static final String KEY_CHANNEL_ID = "channelID";
+    protected static final String SHOW_IN_FOREGROUND = "showInForeground";
```
aurimasc, line 59: Please, namespace the value. Other values are mostly legacy. Going forward, it's a good idea to name values liek "com.unity.xxx".

aurimasc (changes requested):
There are more changes requires (saving&loading, test and test project update), but I can do those myself, since they are complicated.
But please address the issue that breaks backwards compatibility.

com.unity.mobile.notifications/Runtime/Android/Plugins/com/unity/androidnotifications/UnityNotificationManager.java
serefbulbul: Yes makes sense 👍🏻 Updated the value to "com.unity.showInForeground"

com.unity.mobile.notifications/Runtime/Android/AndroidNotification.cs
serefbulbul: Thanks for taking a look @aurimasc. I introduced m_SilentInForeground to back ShowInForeground public interface as you suggested. Please let me know if further changes are necessary

aurimasc:
Extra changes done here: https://github.com/Unity-Technologies/com.unity.mobile.notifications/pull/182

PR #171, opened 2022-04-08
Fix CI for 1.4.x

aurimasc:
Taken Yamato file from master.
Other changes are also mostly taken from master.
iOS tests fail on CI (probably due to provisional auth not set, which can't be set due to our API bug).

### DIFF
--- a/com.unity.mobile.notifications/Tests/Runtime/Android/AndroidNotificationTests.cs
+++ b/com.unity.mobile.notifications/Tests/Runtime/Android/AndroidNotificationTests.cs
@@ -320,7 +320,8 @@ class AndroidNotificationTests
         chOrig.CanShowBadge = true;
         chOrig.EnableLights = true;
         chOrig.EnableVibration = false;
-        chOrig.LockScreenVisibility = LockScreenVisibility.Private;
+        // read-only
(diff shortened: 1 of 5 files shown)


com.unity.mobile.notifications/Tests/Runtime/Android/AndroidNotificationTests.cs
```diff
@@ -10,6 +10,7 @@ class AndroidNotificationTests
     private static int receivedNotificationCount = 0;
 
     [Test]
+    [UnityPlatform(RuntimePlatform.Android)]
```
todi1856, line 13: I think you can have this attribute on top of class, no need to apply it for every test

com.unity.mobile.notifications/Tests/Runtime/iOS/iOSNotificationTests.cs
```diff
@@ -36,6 +76,7 @@ public void AfterEachTest()
     }
 
     [UnityTest]
+    [UnityPlatform(RuntimePlatform.IPhonePlayer)]
```
todi1856, line 79: same as above, can be applied on top of class

todi1856 (approved):

PR #182, opened 2022-04-28
Show in foreground feature for android

aurimasc:


TestProjects/com.unity.mobile-notifications-sample/Assets/Scripts/ScriptableObjects/AndroidNotificationTemplate.cs
```diff
@@ -36,6 +36,7 @@ public class AndroidNotificationTemplate : ScriptableObject
         [TextArea] public string IntentData = "";
         public bool ShowTimestamp = false;
         public long RepeatInterval;
+        public bool ShowInForeground = true;
```
todi1856, line 39: Is technically possible to get notification to show up both while app is in background or foreground. Wondering if this shouldn't be an enum with bit wise values?

com.unity.mobile.notifications/Runtime/Android/AndroidNotification.cs
```diff
@@ -177,11 +177,21 @@ public DateTime CustomTimestamp
             }
         }
 
+        /// <summary>
+        /// Set this notification to be shown when app is in the foreground (default: true).
+        /// </summary>
+        public bool ShowInForeground
+        {
+            get => !m_SilentInForeground;
```
todi1856, line 185: Why the inversion ? I mean why not have m_ShowInForeground, or even have a collapsed property?

com.unity.mobile.notifications/Runtime/Android/Plugins/com/unity/androidnotifications/UnityNotificationUtilities.java
```diff
@@ -29,13 +29,14 @@
...
 import static com.unity.androidnotifications.UnityNotificationManager.KEY_SMALL_ICON;
+import static com.unity.androidnotifications.UnityNotificationManager.KEY_SHOW_IN_FOREGROUND;
 import static com.unity.androidnotifications.UnityNotificationManager.TAG_UNITY;
 
 public class UnityNotificationUtilities {
     // magic stands for "Unity Mobile Notifications Notification"
     private static final byte[] UNITY_MAGIC_NUMBER = new byte[] { 'U', 'M', 'N', 'N'};
     private static final byte[] UNITY_MAGIC_NUMBER_PARCELLED = new byte[] { 'U', 'M', 'N', 'P'};
-    private static final int NOTIFICATION_SERIALIZATION_VERSION = 0;
+    private static final int NOTIFICATION_SERIALIZATION_VERSION = 1;
```
todi1856, line 39: Do you have a test which would show a failure if you would forget to bump this number?

todi1856 (commented):
Questions

TestProjects/com.unity.mobile-notifications-sample/Assets/Scripts/ScriptableObjects/AndroidNotificationTemplate.cs
aurimasc: Notifications in general are designed to be shown while app is in background in order to bring user back to app. But it's possible send them while app is in foreground too.
So ShowInForeground means send notification disregarding foreground/background, because you can safely assume notification should be sent while app is in background.

com.unity.mobile.notifications/Runtime/Android/AndroidNotification.cs
aurimasc: Backward compatibility. This is struct, so if you make one using default, all booleans are false, while this property should default to true, because that was the behavior before it's introduction.

com.unity.mobile.notifications/Runtime/Android/Plugins/com/unity/androidnotifications/UnityNotificationUtilities.java
aurimasc: How do you even write such test? Other than having a expected blob in test.

com.unity.mobile.notifications/Runtime/Android/AndroidNotification.cs
todi1856: This is kind a big inconvenience of struct, what's the reason AndroidNotification is a struct and not a class? Since you might get away this time, but when you'll need a default which is not a "default" of struct, you won't get away with struct.

com.unity.mobile.notifications/Runtime/Android/Plugins/com/unity/androidnotifications/UnityNotificationUtilities.java
todi1856: I had a similar case, I was simply checking the if a specific version corresponds to specific size of serialized data.

I assume here, version 1 produces a specific size of data.

Thus you can simply add a check for failure, for cases where you're serializing more data, but forgot to bump the version. In that case the size will be different than the one you expect for version 1.

com.unity.mobile.notifications/Runtime/Android/AndroidNotification.cs
aurimasc: It's been that since long ago.
It actually makes sense, since it's only contains data.
In any case, it is what it is and changing it to class requires better justification than imaginary "what if in future".

com.unity.mobile.notifications/Runtime/Android/Plugins/com/unity/androidnotifications/UnityNotificationUtilities.java
todi1856: If somebody else picks up this project, and expands serializeNotificationCustom function with extra data, but forgets to update NOTIFICATION_SERIALIZATION_VERSION, then he unknowingly creates a bug. Basically only you know that you need to bump this number, nobody else knows.

None of your current tests will catch such oversight?

aurimasc: Added a comment. If you change the serialization without version check at deserialization, the existing v0 test should blow. Add if (NOTIFICATION_SERIALIZATION_VERSION >...) I hope will be itching eyes enough to go look up what a version is and find that comment :)

todi1856: Still think it's quite easy to make an error. Just consider this:
* I add a new variable in AndroidNotification, newVariable
* I add new line for serializing newVariable (in n serializeNotificationCustom ?)
* I add new line for deserializing newVariable

                if (version > 0)
                {
                    showInForeground = in.readBoolean();
                    newVariable= in.readSomething();
                } 
* I forget to bump NOTIFICATION_SERIALIZATION_VERSION, didn't read the comment

And this where it will fail, assuming previous app install serialized notification data. And no test will warn me, that I did the failure.

From what I can say serializeNotificationCustom_v0 only validates if we correctly deserialize data from version 0, but it doesn't validate the future changes.

todi1856 (approved):
Still think, you're missing an important test for validating the serialization version number. Up to you.

vaidasma:
I see there's branch confilcts in:
com.unity.mobile.notifications/Runtime/Android/Plugins/com/unity/androidnotifications/UnityNotificationManager.java

I've encountered an issue on OS:  5.0.1 device (Samsung Galaxy S4, ) and 5.0.2 ( HTC One M9+), where application crashes on launch

I do not have 5.1 device, therefore not sure if it's affected, but I think this should be addressed.

05/19 05:10:36.912 9203 9281 Debug UnityNotifications Checking for invalid notification IDs still hanging around
05/19 05:10:36.913 9203 9281 Error AndroidRuntime FATAL EXCEPTION: Thread-286
05/19 05:10:36.913 9203 9281 Error AndroidRuntime Process: com.unity3d.mobilenotificationtests, PID: 9203
05/19 05:10:36.913 9203 9281 Error AndroidRuntime java.lang.Error: FATAL EXCEPTION [Thread-286]
05/19 05:10:36.913 9203 9281 Error AndroidRuntime Unity version     : 2020.3.34f1
05/19 05:10:36.913 9203 9281 Error AndroidRuntime Device model      : HTC HTC_M9pw
05/19 05:10:36.913 9203 9281 Error AndroidRuntime Device fingerprint: htc/hiau_ml_tuhl_htc_asia_tw/htc_hiau_ml_tuhl:5.0.2/LRX22G/670935.3:user/release-keys
05/19 05:10:36.913 9203 9281 Error AndroidRuntime Build Type        : Development
05/19 05:10:36.913 9203 9281 Error AndroidRuntime Scripting Backend : Mono
05/19 05:10:36.913 9203 9281 Error AndroidRuntime ABI               : armeabi-v7a
05/19 05:10:36.913 9203 9281 Error AndroidRuntime Strip Engine Code : false
05/19 05:10:36.913 9203 9281 Error AndroidRuntime 
05/19 05:10:36.913 9203 9281 Error AndroidRuntime Caused by: java.lang.NoSuchMethodError: No virtual method getActiveNotifications()[Landroid/service/notification/StatusBarNotification; in class Landroid/app/NotificationManager; or its super classes (declaration of 'android.app.NotificationManager' appears in /system/framework/framework.jar)
05/19 05:10:36.913 9203 9281 Error AndroidRuntime 	at com.unity.androidnotifications.UnityNotificationManager.findInvalidNotificationIds(UnityNotificationManager.java:434)
05/19 05:10:36.913 9203 9281 Error AndroidRuntime 	at com.unity.androidnotifications.UnityNotificationManager.performNotificationHousekeeping(UnityNotificationManager.java:407)
05/19 05:10:36.913 9203 9281 Error AndroidRuntime 	at com.unity.androidnotifications.UnityNotificationManager.lambda$triggerHousekeeping$0(UnityNotificationManager.java:391)
05/19 05:10:36.913 9203 9281 Error AndroidRuntime 	at com.unity.androidnotifications.-$$Lambda$UnityNotificationManager$fSttc7CuZmP-kDH-q_FquZSJ7Yk.run(lambda)
05/19 05:10:36.913 9203 9281 Error AndroidRuntime 	at java.lang.Thread.run(Thread.java:818)





Tested on devices:
• VLNQA00140 - Samsung Galaxy S4 (GT-I9505), CPU: Snapdragon S4 Plus MSM8960, GPU: Adreno 320, OS: 5.0.1
• VLNQA00057 - HTC One M9+ (HTC_M9pw), CPU: MediaTek Helio X10 MT6795T, GPU: PowerVR Rogue G6200, OS: 5.0.2

vaidasma (approved):
Good from my side:
• Rechecked OS 5 devices. Application is now able to run
• On the OS:5 devices now also works as intended: notifications, not meant to appear while app is in foreground, now is proven to behave so

image

PR #187, opened 2022-05-09
Android/jni refactor and optimization

aurimasc:
The thing in this PR is performance. JNI calls are expensive and scheduling a notification is way too costly at the moment.
This PR addresses the C#<->Java interop part of the problem. Setting up Notification.Builder takes milliseconds of time (low single digit figure, but that's still huge time). The main problem is that when using AndroidJavaObject we actually invoke Java code where reflection is used to find the method to invoke (there is some cashing there). In latest Unity we have improvement in the area where we can call method via jmethodID, a significant shortcut.
Tested on 3 different devices and in all cases I see a fairly consistent ~3x improvement for creating and setting up Notification.Builder on 22.2 (didn't measure older versions, but on average should also be better due to not calling Java for no reason, like to set the value that is already the default).

Testing done:
- Automated test test sending/receiving and all parameter setting/getting
- Performance tested using specially made project on Nokia 7 Plus (10.0), Asus ROG (8.1) and Samsung Galaxy S7 (7.0)
- On first two also ran automated tests locally using 2022.2

Request for QA:
- the key split is Unity 2022.2 vs older versions (different implementation)
- make sure notification parameters work, use fancy notifications (colors, icons); in package project should have enough buttons for that
- no need to test many Unity versions, one version older then 2022.2 on one or two devices should suffice for "older"
- Unity 2022.2 needs more testing, but there's a split:
  - test the same fancy notifications in this version on one or two devices
  - we do collect most of required stuff during initialization and blow up if something is not found; send one notification on device and if it works, you can assume it's OK; try this on more devices and different android versions

### DIFF
--- a/com.unity.mobile.notifications/Runtime/Android/AndroidNotificationCallback.cs
+++ b/com.unity.mobile.notifications/Runtime/Android/AndroidNotificationCallback.cs
@@ -1,3 +1,4 @@
+using System;
 using UnityEngine;
 
 namespace Unity.Notifications.Android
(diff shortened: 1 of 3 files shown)


todi1856 (approved):

vaidasma:
Verified as fixed.

I've not encountered anything that seemed like a fauty behaviour:
• Tested different types of notifications and their build speed
• Used various Android devices to see if notifications arrive, are not slow to be created, and visually appear as they should

image

vaidasma (approved):

vaidasma:
Also tested on runs with 100 notifications being queued. In all cases, the changes made notification building quicker by a sizeable amount of time:

image


image

PR #197, opened 2022-05-27
Merge manager classes

aurimasc:
Merge UnityNotificationManager and UnityNotificationManagerOreo classes into one.
We had two classes where UnityNotificationManager would provide a simulation implementation of channels, while UnityNotificationManagerOreo would override the same methods and implement real channel support. This causes more confusion than is worth, since channel support code is not that big.
Also, the current state of UnityNotificationManager where we actually have two instances and some statics is difficult to maintain and is error prone. The intent is to move away from that and two classes are a bit in a way, so another point for merging them.

Additionally move NotificationChannelWrapper and NotificationCallback into the same source file, they are very small, no real gain with that.
Lastly, querying channels was a bit optimized to avoid excessive allocations, for example return null when there are no channels and save C# size from doing several expensive JNI calls to figure that out. Added tests to cover those.

Tested on Nokia 7 Plus (10.0) and Samsung Galaxy S7 (7.0) as well as ran channel related tests on both to make sure both Oreo+ and pre-O code paths work.


com.unity.mobile.notifications/Runtime/Android/Plugins/com/unity/androidnotifications/UnityNotificationManager.java
```diff
@@ -931,3 +972,23 @@ private static synchronized Notification removeScheduledNotification(Integer id)
         return mScheduledNotifications.remove(id);
     }
 }
+
+// Provide a wrapper for NotificationChannel.
+// Create this wrapper for all Android versions as NotificationChannel is only available for Android O or above.
+class NotificationChannelWrapper {
```
todi1856, line 978: Why move NotificationChannelWrapper & NotificationChannelWrapper out of their respective files, putting multiple classes in one file makes things less readable.

Couldn't you remove public keyword in those files instead?

com.unity.mobile.notifications/Runtime/Android/Plugins/com/unity/androidnotifications/UnityNotificationManager.java
```diff
@@ -220,44 +228,77 @@ protected static NotificationChannelWrapper getNotificationChannel(Context conte
...
         return UnityNotificationManager.getNotificationChannel(mContext, id);
     }
 
-    // Delete a notification channel by id.
-    // This function will only be called for devices which are low than Android O.
     public void deleteNotificationChannel(String id) {
-        SharedPreferences prefs = mContext.getSharedPreferences(NOTIFICATION_CHANNELS_SHARED_PREFS, Context.MODE_PRIVATE);
-        Set<String> channelIds = new HashSet<String>(prefs.getStringSet(NOTIFICATION_CHANNELS_SHARED_PREFS_KEY, new HashSet<String>()));
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            getNotificationManager().deleteNotificationChannel(id);
```
todi1856, line 257: could have added return here, thus not needing to move stuff in else block - would produce less changes

com.unity.mobile.notifications/Runtime/Android/Plugins/com/unity/androidnotifications/UnityNotificationManager.java
```diff
@@ -220,44 +228,77 @@ protected static NotificationChannelWrapper getNotificationChannel(Context conte
...
+            int i = 0;
+            for (NotificationChannel ch : channels) {
+                channelList[i++] = notificationChannelToWrapper(ch);
+            }
 
-        for (String k : channelIdsSet) {
-            channels.add(getNotificationChannel(k));
+            return channelList;
+        } else {
+            SharedPreferences prefs = mContext.getSharedPreferences(NOTIFICATION_CHANNELS_SHARED_PREFS, Context.MODE_PRIVATE);
```
todi1856, line 291: no need to put stuff in else block

todi1856 (approved):
minor comments

com.unity.mobile.notifications/Runtime/Android/Plugins/com/unity/androidnotifications/UnityNotificationManager.java
aurimasc: Just removing public would do it too, but the two are very small and specific, it feels very excessive to have dedicated files for them. Now that the manager classes are merged, they are used only in this Java class, kind of should be nested classes actually.

com.unity.mobile.notifications/Runtime/Android/Plugins/com/unity/androidnotifications/UnityNotificationManager.java
todi1856: I guess we won't agree here, but I'll just say this doesn't scale when you have multiple if/else inside

aurimasc: I'd say the opposite.
This is the situation where you have strict this-or-that. By having if-else and no return after it, then compiler will require both sides of if to return a value in all possible cases. Without else a missing return in then part leads to supposed else part being executed, which is wrong.

PR #206, opened 2022-06-27
Fix iOS tests on trunk & few minor changes

aurimasc:
Changes:
- Switch iOS agent to fix running tests in trunk
- Fix package description: wrong Android version; remove that entirely, since the correct thing is in top page of the documentation, better mention push notifications instead and not have one more place to fix when bumping


todi1856 (approved):
Please consider adding Petras or Pavel instead of me for iOS stuff in the future

Issue #202, opened 2022-06-13
[Feature Request] Notification without sound 


hanilozmen:
I believe that it is generally best practice to send push notifications without sound for regular notifications. It is achievable in android by arranging push notification priority(low priority in general). However, it is not available in ios currently( v.2.0.2).

After a little investigation, I came up with a workaround in ios : 
In this file , _com.unity.mobile.notifications/Runtime/iOS/Plugins/UnityNotificationManager.m_ 
I commented out the line ( line 301) for this purpose:
` //content.sound = [UNNotificationSound defaultSound];`

I think that it would be nice to add a bool parameter to _Unity.Notifications.iOS.iOSNotification_ class in the library.

aurimasc:
Feature implemented.

PR #213, opened 2022-08-05
Add sound type support for iOS

aurimasc:
User requested an ability to have notification without sound: https://github.com/Unity-Technologies/com.unity.mobile.notifications/issues/202

This PR is exposing entire set of sound APIs:
- Choice of type: None, Default, Critical, Ringtone
- Ability to specify custom sound
- Sound volume (for critical)

Note, that critical alerts require special entitlement from Apple, while ringtone is not documented. Just exposing them in API, in general this PR concerns silent notifications and custom sound.
One of existing notifications in test project was changed to have no sound.
Custom sound support is tricky since our API to manipulate XCode project doesn't look complete. Basically you need to specify the file name in the api and add that file to XCode project manually (right click in the tree and select add). For now lets have it this way, we could add better support later, if anyone asks for it.

I've tried the feature myself, but one notable thing is that some features are iOS version specific (see changes to UnityNotificationManager.m), so default should be used as fallback on device with older versions (not blow up).


todi1856 (approved):

vaidasma (approved):
Exposed API can be used.

Custom sounds can be used with Deault notification sound type

Issue #211, opened 2022-07-26
[Android] OutOfMemoryError still persists in version 2.0.2


tomurtogu:
The issue I had reported for version 2.0.1 still persists in version 2.0.2

Unity version: 2020.3.34f
Devices
Redmi Note 10 (Android 12)
Nokia G10 (Android 12)

java.lang.OutOfMemoryError: 
  at android.os.Parcel.createString8Array (Parcel.java:1477)
  at android.content.pm.ApplicationInfo.<init> (ApplicationInfo.java:2019)
  at android.content.pm.ApplicationInfo.<init> (ApplicationInfo.java:62)
  at android.content.pm.ApplicationInfo$1.lambda$createFromParcel$0 (ApplicationInfo.java:1988)
  at android.content.pm.ApplicationInfo$1$$ExternalSyntheticLambda0.readRawParceled (Unknown Source)
  at android.os.Parcel.readSquashed (Parcel.java:2122)
  at android.content.pm.ApplicationInfo$1.createFromParcel (ApplicationInfo.java:1988)
  at android.content.pm.ApplicationInfo$1.createFromParcel (ApplicationInfo.java:1985)
  at android.os.Parcel.readParcelable (Parcel.java:3333)
  at android.os.Parcel.readValue (Parcel.java:3226)
  at android.os.Parcel.readArrayMapInternal (Parcel.java:3623)
  at android.os.BaseBundle.initializeFromParcelLocked (BaseBundle.java:292)
  at android.os.BaseBundle.unparcel (BaseBundle.java:236)
  at android.os.Bundle.getParcelable (Bundle.java:1002)
  at android.app.Notification$Builder.recoverBuilder (Notification.java:6397)
  at com.unity.androidnotifications.UnityNotificationUtilities.recoverBuilder (UnityNotificationUtilities.java:517)
  at com.unity.androidnotifications.UnityNotificationManager.loadSavedNotifications (UnityNotificationManager.java:519)
  at com.unity.androidnotifications.UnityNotificationRestartOnBootReceiver.rescheduleSavedNotifications (UnityNotificationRestartOnBootReceiver.java:26)
  at com.unity.androidnotifications.UnityNotificationRestartOnBootReceiver.onReceive (UnityNotificationRestartOnBootReceiver.java:21)
  at android.app.ActivityThread.handleReceiver (ActivityThread.java:4465)
  at android.app.ActivityThread.access$1800 (ActivityThread.java:275)
[...]

Issue #210, opened 2022-07-22
android.content.res.Resources$NotFoundException


korober:
I still have a lot of crashes in Google Console:
Unity: 2021.3.5f1
Mobile Notifications: 2.0.2 (May 19, 2022) - latest release version in upm
Additional plugins: Firebase Unity Sdk: 9.20
[Here is a lot ](https://forum.unity.com/threads/mobile-notification-package.620227/page-15)of messages about the issue
```
java.lang.RuntimeException: 
  at android.app.ActivityThread.handleReceiver (ActivityThread.java:3107)
  at android.app.ActivityThread.-wrap18 (ActivityThread.java)
  at android.app.ActivityThread$H.handleMessage (ActivityThread.java:1611)
  at android.os.Handler.dispatchMessage (Handler.java:110)
  at android.os.Looper.loop (Looper.java:203)
  at android.app.ActivityThread.main (ActivityThread.java:6251)
  at java.lang.reflect.Method.invoke (Native Method)
  at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run (ZygoteInit.java:1063)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:924)
Caused by: android.content.res.Resources$NotFoundException: 
  at android.app.ResourcesManager.createAssetManager (ResourcesManager.java:253)
  at android.app.ResourcesManager.createResourcesImpl (ResourcesManager.java:310)
  at android.app.ResourcesManager.getOrCreateResources (ResourcesManager.java:571)
  at android.app.ResourcesManager.getResources (ResourcesManager.java:645)
  at android.app.ActivityThread.getTopLevelResources (ActivityThread.java:1920)
  at android.app.LoadedApk.getResources (LoadedApk.java:771)
  at android.app.ContextImpl.<init> (ContextImpl.java:2055)
  at android.app.ContextImpl.createApplicationContext (ContextImpl.java:1824)
  at android.content.ContextWrapper.createApplicationContext (ContextWrapper.java:792)
  at android.content.ContextWrapper.createApplicationContext (ContextWrapper.java:792)
  at android.app.Notification$Builder.recoverBuilder (Notification.java:3803)
  at com.unity.androidnotifications.UnityNotificationUtilities.recoverBuilder
  at com.unity.androidnotifications.UnityNotificationManager.loadSavedNotifications (UnityNotificationManager.java)
[...]

Issue #208, opened 2022-07-06
Failed to serialize Parcelable error on notification with emoji


sadegh-askari:
Hi,
I tried to use emoji in text of my notifications. I created two notifications with emoji and with different delivery times.
The first time when the program is closed, the notification is displayed with emoji without any problem, but the next time I enter the program and the notifications set, when I want to close the app, these errors are displayed and notifications aren't delivered.

> 2022-07-06 13:29:15.747 29147-29219/? E/UnityNotifications: Failed to serialize Parcelable
    java.lang.RuntimeException: Tried to marshall a Parcel that contained Binder objects.
        at android.os.Parcel.nativeMarshall(Native Method)
        at android.os.Parcel.marshall(Parcel.java:480)
        at com.unity.androidnotifications.UnityNotificationUtilities.serializeParcelable(UnityNotificationUtilities.java:186)
        at com.unity.androidnotifications.UnityNotificationUtilities.serializeNotificationParcel(UnityNotificationUtilities.java:106)
        at com.unity.androidnotifications.UnityNotificationUtilities.serializeNotification(UnityNotificationUtilities.java:87)
        at com.unity.androidnotifications.UnityNotificationManager.saveNotification(UnityNotificationManager.java:496)
        at com.unity.androidnotifications.UnityNotificationManager.scheduleNotification(UnityNotificationManager.java:278)
        at com.unity3d.player.UnityPlayer.nativeFocusChanged(Native Method)
        at com.unity3d.player.UnityPlayer.access$000(Unknown Source:0)
        at com.unity3d.player.UnityPlayer$e$1.handleMessage(Unknown Source:179)
        at android.os.Handler.dispatchMessage(Handler.java:101)
        at android.os.Looper.loop(Looper.java:164)
        at com.unity3d.player.UnityPlayer$e.run(Unknown Source:20)

> 2022-07-06 13:29:43.275 29336-29336/? E/UnityNotifications: Failed to deserialize notification
    java.lang.NullPointerException: Attempt to invoke virtual method 'android.app.Notification$Builder com.unity.androidnotifications.UnityNotificationManager.createNotificationBuilder(java.lang.String)' on a null object reference
        at com.unity.androidnotifications.UnityNotificationUtilities.deserializeNotificationCustom(UnityNotificationUtilities.java:325)
        at com.unity.androidnotifications.UnityNotificationUtilities.deserializeNotification(UnityNotificationUtilities.java:223)
[...]

PR #215, opened 2022-08-11
Rework serialization

aurimasc:
Change the way we save pending notifications.
Users can send notification pure by using our API, or they can access the Notification.Builder and customize it using Android APIs.
This PRs feeds the information about it to serialization code, so if we know notification was not customized, we use the simple serialization system, since that one is more reliable with regard to app or OS upgrades, transfer to new device etc. Of course, notifications that might have been customized have to be saved as is and we have to be ready for failures when deserializing.
One other change is introduction of separate Java file for code that is only used by tests. This file is only used when running tests, not included in normal build to not bloat the app size.

The system is in general covered by automated test, so the manual testing is required for:
- schedule notification & kill the app, to see if notification is still delivered
- reschedule after restart
- try several devices and most importantly different android version

This rework should in general address these issues (we already handle the exceptions, but after this PR most of them will be prevented from happening in the first place):
https://github.com/Unity-Technologies/com.unity.mobile.notifications/issues/211
https://github.com/Unity-Technologies/com.unity.mobile.notifications/issues/210
https://github.com/Unity-Technologies/com.unity.mobile.notifications/issues/208


com.unity.mobile.notifications/Runtime/Android/Plugins/com/unity/androidnotifications/UnityNotificationUtilities.java
```diff
@@ -72,43 +72,44 @@ protected static int findResourceIdInContextByName(Context context, String name)
...
-                fallback = Base64.encodeToString(bytes, 0, bytes.length, 0);
+            if (serializeParcel) {
+                Intent intent = new Intent();
+                intent.putExtra(KEY_NOTIFICATION, notification);
+                if (serializeNotificationParcel(intent, out)) {
+                    out.close();
+                    byte[] bytes = data.toByteArray();
+                    serialized = Base64.encodeToString(bytes, 0, bytes.length, 0);
+                } else {
+                    return; // failed
```
todi1856, line 91: shouldn't we log it, like you do below Log.e(TAG_UNITY, "Failed to serialize notification", e); ?

com.unity.mobile.notifications/Runtime/Android/Plugins/com/unity/androidnotifications/UnityNotificationUtilities.java
```diff
@@ -72,43 +72,44 @@ protected static int findResourceIdInContextByName(Context context, String name)
...
-                out.close();
-                byte[] bytes = data.toByteArray();
-                serialized = Base64.encodeToString(bytes, 0, bytes.length, 0);
+            else {
+                if (serializeNotificationCustom(notification, out)) {
+                    out.flush();
+                    byte[] bytes = data.toByteArray();
+                    serialized = Base64.encodeToString(bytes, 0, bytes.length, 0);
+                } else {
+                    return; // failed
```
todi1856, line 100: silent error?

com.unity.mobile.notifications/Tests/Runtime/Android/AndroidNotificationSimpleTests.cs
```diff
@@ -3,11 +3,43 @@
...
+    : IPrebuildSetup, IPostBuildCleanup
 {
     const string kChannelId = "SerializeDeserializeNotificationChannel";
 
+#if UNITY_EDITOR
+    PluginImporter GetTestUtils()
+    {
+        var assets = AssetDatabase.FindAssets("UnityNotificationTestUtils");
+        if (assets.Length != 1)
+            throw new Exception("Something is wrong");
```
todi1856, line 20: Not a very good error, why not say 'Error while acquiring UnityNotificationTestUtils asset' or smth?

com.unity.mobile.notifications/Tests/Runtime/Android/AndroidNotificationSimpleTests.cs
```diff
@@ -3,11 +3,43 @@
...
+        return (PluginImporter)AssetImporter.GetAtPath(AssetDatabase.GUIDToAssetPath(assets[0]));
+    }
+#endif
+
+    public void Setup()
+    {
+#if UNITY_EDITOR
+        var testUtils = GetTestUtils();
+        testUtils.SetCompatibleWithPlatform(BuildTarget.Android, true);
```
todi1856, line 29: can you explain why this is needed, meaning why UnityNotificationTestUtils is not commited to be always compatible with Android?

todi1856 (approved):
I think looks fine, few questions

com.unity.mobile.notifications/Runtime/Android/Plugins/com/unity/androidnotifications/UnityNotificationUtilities.java
aurimasc: serializeNotificationParcel already logs it, so this would be a second logging for the same error.

com.unity.mobile.notifications/Runtime/Android/Plugins/com/unity/androidnotifications/UnityNotificationUtilities.java
aurimasc: serializeNotificationCustom logs the error, so no need for one more

com.unity.mobile.notifications/Tests/Runtime/Android/AndroidNotificationSimpleTests.cs
aurimasc: Because then it is always copied into built player, while we only need it for running tests.

vaidasma (approved):
Verified.

• Notifications are still delivered in different app states: killed, in foreground, in background, after restart (Exception: Xiaomi Mi 5s, known not to receive notifications after restart)

Tested on devices:
• VLNQA00090 - Samsung Galaxy S8 (USA) (SM-G950U), CPU: Snapdragon 835 MSM8998, GPU: Adreno 540, OS: 7.0.0
• VLNQA00341 - Note20 Ultra 5G INT (SM-N986B), CPU: Exynos 990, GPU: Mali-G77, OS: 10

Could not test on:

Issue discovered (unrelated to PR): Notifications do not arrive until app is in foreground
Issue link: [MPT-1819](https://jira.unity3d.com/browse/MPT-1819)

PR #216, opened 2022-08-17
Refactor notification manager

aurimasc:
The problem this PR is solving is that UnityNotificationManager class is used as the central manager for most notification stuff and as a receiver for alarms when notifications are supposed to be shown. The receiver objects are created by OS and they are recreated based on vairous conditions, while our own manager is supposed to be singleton instance. Some issues coming from that were solved via static variables in the class, which was error prone.
The changes in this PR are:
- have a single static instance of the manager, all other static variables converted to instance variables in this singleton
- receivers forward messages to the singleton instance
- a bunch of cleanup related to method visibility and reduction of static methods
- some minor cleanup

Note to QA:
- Android only
- various notification types shouldn't be affected, only scheduling and receiving
- testing required to check that scheduled notifications are received in all supported cases: app open, in background, killed, after device reboot (the first one is covered by automated tests and is least risky, can be skipped)
- essential to tests a range of supported OS versions
- tested by me on Nokia 7 Plus (10.0)

### DIFF
--- a/com.unity.mobile.notifications/Runtime/Android/Plugins/com/unity/androidnotifications/UnityNotificationManager.java
+++ b/com.unity.mobile.notifications/Runtime/Android/Plugins/com/unity/androidnotifications/UnityNotificationManager.java
@@ -52,24 +52,24 @@ public class UnityNotificationManager extends BroadcastReceiver {
     protected UnityNotificationBackgroundThread mBackgroundThread;
     protected Random mRandom;
 
-    protected static final String TAG_UNITY = "UnityNotifications";
+    static final String TAG_UNITY = "UnityNotifications";
 
-    protected static final String KEY_FIRE_TIME = "fireTime";
-    protected static final String KEY_ID = "id";
-    protected static final String KEY_INTENT_DATA = "data";
-    protected static final String KEY_LARGE_ICON = "largeIcon";
-    protected static final String KEY_REPEAT_INTERVAL = "repeatInterval";
-    protected static final String KEY_NOTIFICATION = "unityNotification";
-    protected static final String KEY_NOTIFICATION_ID = "com.unity.NotificationID";
-    protected static final String KEY_SMALL_ICON = "smallIcon";
-    protected static final String KEY_CHANNEL_ID = "channelID";
-    protected static final String KEY_SHOW_IN_FOREGROUND = "com.unity.showInForeground";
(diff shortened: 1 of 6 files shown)


com.unity.mobile.notifications/Runtime/Android/Plugins/com/unity/androidnotifications/UnityNotificationManager.java
```diff
@@ -111,30 +111,30 @@ public UnityNotificationManager(Context context, Activity activity) {
...
-        return getNotificationManagerImpl(context, (Activity) context);
+    static synchronized UnityNotificationManager getNotificationManagerImpl(Context context) {
+        if (mUnityNotificationManager == null) {
+            mUnityNotificationManager = new UnityNotificationManager();
+            mUnityNotificationManager.mVisibleNotifications = new HashSet<>();
+            mUnityNotificationManager.mScheduledNotifications = new ConcurrentHashMap();
+        }
+
+        // always assign context, as it might change
+        mUnityNotificationManager.mContext = context.getApplicationContext();
```
todi1856, line 122: Looks a bit unsafe, wouldn't it be better for mUnityNotificationManager to call context.getApplicationContext() when returning context instead of caching it?

todi1856 (approved):
Minor comment

com.unity.mobile.notifications/Runtime/Android/Plugins/com/unity/androidnotifications/UnityNotificationManager.java
aurimasc: We've been caching it all the time, so it's quite safe. Here I'm considering scenarios like app update.

vaidasma:
Verified.

• Notifications are still delivered in different app states: killed, in foreground, in background, after restart (Exception: Xiaomi Mi 5s, known not to receive notifications after restart)

Tested on devices:
• VLNQA00090 - Samsung Galaxy S8 (USA) (SM-G950U), CPU: Snapdragon 835 MSM8998, GPU: Adreno 540, OS: 7.0.0
• VLNQA00341 - Note20 Ultra 5G INT (SM-N986B), CPU: Exynos 990, GPU: Mali-G77, OS: 10

Could not test on:

Issue discovered (unrelated to PR): Notifications do not arrive until app is in foreground
Issue link: [MPT-1819](https://jira.unity3d.com/browse/MPT-1819)

vaidasma (approved):

PR #218, opened 2022-09-01
iOS Location improvements and fixes

aurimasc:
* After test project update for Android, it got broken for iOS. This PR fixes it.
* We used Vector2 to specify the location in location trigger, which has a couple of problems:
  - Uses floats while Apple API uses doubles - risk of precision loss (with number being degrees can be meaningful)
  - Sort-of swapped with X being latitude and Y being longitude; looking at map you'd expect the opposite
  This PR introduces proper properties for Latitude and Longitude
* Apple API has repeat capability for location trigger, we didn't expose it, now we do
* The location button in test project was changed to be repeatable and work in foreground for easier testing

QA:
* iOS only changes
* Stuff already tested, no real need to test, but probably worth trying out
* No automated test possible ATM due to testing framework issues (same problem as with push tests)
* I could not get location simulation working on Simulator
* To test on real device:
  - Use Debug -> Simulate Location in Xcode
  - Add the .gpx file to Xcode project (available in test project, use Debug -> Simulate Location -> Add GPS... in Xcode); when done, a new item will appear in the same menu
  - Looks like you need to simulate some location twice to get "enterred" event: first some location outside, then a desired location; for simplicity best to have a repeatable trigger with both enter and leave events, then simply jump around the locations.

### DIFF
--- a/TestProjects/Main/Assets/Resources/iOSNotifications/LocationTrigger/Location Triggered.asset
+++ b/TestProjects/Main/Assets/Resources/iOSNotifications/LocationTrigger/Location Triggered.asset
@@ -13,18 +13,19 @@ MonoBehaviour:
   m_Name: Location Triggered
   m_EditorClassIdentifier: 
   ButtonName: Send A Notification (Location Trigger)
-  Identifier: 
+  Identifier: loc_not
   CategoryIdentifier: category_indentifier
   ThreadIdentifier: thread_identifier
   Title: Location Notification Title
   Subtitle: This is a subtitle
   Body: And this is the body of this notification
-  ShowInForeground: 0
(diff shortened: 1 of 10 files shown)


vaidasma:
Unable to trigger the location notifications by simulating GPS coordinates via Xcode - queued notification does not trigger. Will be contacting the developer

vaidasma:
Location triggers can be received (while using the .gpx file as well as predefined Xcode locations):

Tested with devices
• VLNQA00237 - iPhone 6s (MKQJ2VC/A), CPU: A9, GPU: PowerVR GT7600, OS: 14.3

vaidasma (approved):

PR #220, opened 2022-09-06
Android 12+ exact scheduling

aurimasc:
https://jira.unity3d.com/browse/MNB-10
Historically notifications were scheduled using exact scheduling on Android 6 and newer, older versions only supported inexact scheduling (notification would appear around the requested time, but could easily be a minute early or late). Sin Android 12 Google is pushing towards inexact with new permissions and control. Due to SDK being shipped we only supported inexact scheduling on Android 12.
Changes in this PR:
- Adds new setting to disable exact scheduling as well as add the permissions
- Implements exact scheduling support for Android 12+; has to use reflection, because older Unity versions are shipped with older SDK, would be terrible experience to require user to manually install newer
- Also implements exact/inexact based on the setting
- Some constant refactoring (not actually related to the rest or PR)

Testing:
- Tested on Samsung S10e (12.0) that new functionality works
- Additional testing required:
  - when SCHEDULE_EXACT_ALARM permission is used, on Android 12+ exact scheduling can be controlled in Settings (Alarms & Reminders), should work in both cases (just exactness as requested)
  - when USE_EXACT_ALARM permission is used, on Android 13 exact scheduling should work normally;
  - with both permissions, USE_EXACT_ALARM applies to Android 13, while SCHEDULE_EXACT_ALARM applies to Android 12
  - I had no Android 13 device to actually test, also note this: https://jira.unity3d.com/browse/MNB-13 ; in case with target SDK 33 it's not possible to test on Android 13, suspend this till that is implemented
  - Reschedule after reboot was minorly touched by refactoring, try on few devices


vaidasma:
The settings can be a bit misleading for the users, including having the boolean list. Is there cases where these are not used together? Can it not be joined together?




image


Leaving PR on hold as Android 13 notifications are still not able to be delivered


Tested:
• Schedulling after restart
• Schedule at the exact time: everything / nothing

Pixel 6 is unable to receive notifications to the notification bar

Tested on devices so far:
• VLNQA00267 - Samsung Galaxy S10+ (SM-G975F), CPU: Exynos 9 9820, GPU: Mali-G76, OS: 11

vaidasma (approved):
Notifications are now received while app is in foreground with Android 13 devices

PR #221, opened 2022-09-14
Android API 33 support

aurimasc:
https://jira.unity3d.com/browse/MNB-13
Android 13 addes a new runtime permission for posting notifications. If target SDK is less than 33, OS will ask for permission on first launch, while for 33 and later the request has to be made from code.
Changes by this PR:
- add permission to manifest (otherwise notifications don't work at all)
- introduce new APIs for checking and asking for permission to post notifications
- new button in test project to request permission
- minor unrelated thread safety issue in Java, that was noticed

Here's how it works:
- On devices with older Android versions nothing changes, new APIs do nothing and report that notifications are allowed.
- On Android 13 permission needs to be requested, otherwise notifications work silently (no UI show up, but callbacks are still received, just like iOS Provisional authorization)
- Permission can be denied with "Don't ask again", on Pixel 5 this is done by denying twice; when this is done, no permission is asked again; uninstall the app to reset
- Permission can be given in settings (test projects Open Settings button gets you there); when this is used, it is allow or deny don't ask for the app, this can be used to overrule earlier allow/permanent deny
- Permission has to requested by app if both device and application targetSDK are 33, if targetSDK is lower, permission is asked for automatically on first launch and app cannot ask for it at runtime

Tested on devices:
- Asus ROG Phone (8.1)
- Pixel 5 (13)

No special requests for testing, just double check based on info above.

### DIFF
--- a/TestProjects/Main/Assets/Scripts/AndroidTest.cs
+++ b/TestProjects/Main/Assets/Scripts/AndroidTest.cs
@@ -93,6 +93,8 @@ namespace Unity.Notifications.Tests.Sample
             ((Action)m_groups["Channels"]["Create Secondary Simple Channel"]).Invoke();
             ((Action)m_groups["Channels"]["Create Fancy Channel"]).Invoke();
             m_LOGGER.Clear().White("Welcome!");
+
+            HandleNotificationPermission();
             HandleLastNotificationIntent();
         }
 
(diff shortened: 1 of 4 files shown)


todi1856 (approved):

vaidasma (approved):
Notifications can now be received when app is in background (after successfully requesting the permission for the notifications)

Tested with device:
VLNQA00460 - Google Pixel 6 (Pixel 6), CPU: Google Tensor (Whitechapel), GPU: Mali-G78, OS: 13

PR #225, opened 2022-09-26
Backport Android 12+ fixes to 1.4.x series

aurimasc:
We still support 1.4.x notification version as it is compatible with API levels that are minimum on 2020.3.
The changes backported are for exact scheduling and notification permission.

To keep changes to minimum, not doing anything related to scheduling related permissions, just document that one of the two is required to support exact scheduling on Android 12+. User has to add it manually.
Add POST_NOTIFICATIONS to manifest. Not adding permission request API, just documenting how to request it.

QA: the only real change here is that Android 13 devices work fine with target SDK less than 33 and with 33 (in the later case one has to manually request the permission, test project has button for that).

### DIFF
--- a/TestProjects/com.unity.mobile-notifications-sample/Scripts/AndroidTest.cs
+++ b/TestProjects/com.unity.mobile-notifications-sample/Scripts/AndroidTest.cs
@@ -4,6 +4,7 @@ using System.Collections;
 using System.Collections.Generic;
 using System.Collections.Specialized;
 using UnityEngine;
+using UnityEngine.Android;
 using UnityEngine.UI;
 
 #if UNITY_ANDROID || UNITY_EDITOR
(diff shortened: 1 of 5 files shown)


todi1856 (approved):

vaidasma (approved):
Verified as fixed
• Android 13 device can successfully send / receive Notifications at Target API level 33 (after requesting permission)
• Android 13 device can successfully send / receive Notifications at Target API level 28 (permission request prompted at app launch)
• Android 13 device can successfully send / receive Notifications at Target API level 22 (permission request prompted at app installation)


Tested on:
VLNQA00460 - Google Pixel 6 (Pixel 6), CPU: Google Tensor (Whitechapel), GPU: Mali-G78, OS: 13

PR #229, opened 2022-10-10
Use Color32 for textures

smitdylan2001:
I rewrote texture utils to use Color32 instead of color. Because of this the feature should use less memory


aurimasc:
What is the point of the change?
This is a feature that is used in Editor only and it is not the one that is called frequently. So I don't think memory usage here is of any concern, while losing image quality is concerning.

Issue #194, opened 2022-05-23
[Feature Request] Please add BigPictureStyle Notification (For Android)


wony855:
Hello~

I tried to implement BigPictureStyle Notification using builder but failed...
I hope you officially support this feature

Thank you

sagardaingade:
I am waiting for this feature.

yenmoc:
i'm waiting too

aurimasc:
Implemented in master.

PR #235, opened 2022-11-16
Android add support for big picture style

aurimasc:
https://jira.unity3d.com/browse/MNB-18
https://github.com/Unity-Technologies/com.unity.mobile.notifications/issues/194

Add support for BigPicture style notifications. It's a predefined notification style where main thing in notification is an image.

Main thing: we add new APIs that expose to users this Android API: https://developer.android.com/reference/android/app/Notification.BigPictureStyle#bigLargeIcon(android.graphics.Bitmap)
The availability of individual properties depend on Android version, see Android docs for that.

Because just being able to show an image from apps resources feel too limited fur such kind of app, I made it to support local files and URIs (for the later we only support what Android does and it's do it your self thing when it comes to actually serving images). The URIs support have been also added for small and large icons of the standard notification support.

To facilitate testing a separate test project created just for that. Serving content from app is not so trivial and it was easiest to use AndroidX library for that, which I did not want to put in main test project, as we also use it for running tests.

Testing required:
- Use the new dedicated project; note that it requires SDK 33, if you have problems, easiest way is to export Android Studio project and setup SDK there (I had problems with 20.3 installed via hub)
- BigPicture is supported on all Android versions we support, but the level of support differs. Looks like URIs are supported since Android 6 and a couple new features were added in Android 12. Best to go to Android docs linked above.
- Use the slider in test project to set a delay; test notifications with and without killing the app after sending.
- One of the buttons lets you choose the image from pictures, make some photo in advance. Note, that I had problems on Pixel 5 where not all images could be used (when selecting image a small error popup was shown), so setup an image in local storage for that.
- Try making a project that downloads image from the internet and shows it in notification. To see if adding a code sample to manual is paramount for that :)
- Important to test on a range of Android versions, since even for the same functionality different versions take different code paths (some new, more efficient ways added in newer Androids).
- Devices used by me (using the AndroidPigPicture project that is added by this PR):
  - Asus ROG (8.1)
  - Nokia 7 Plus (10.0)
  - Pixel 5 (13)
 
Notes:
- On Pixel 5 (13) I don't see ContentDescription doing anything visual to notification, while other devices don't support this feature.


todi1856 (approved):

todi1856 (approved):
P.S I think the changes should be always accompanied by entry in CHANGELOG, it's wrong to add entries in CHANGELOG later

guobyt (approved):
Testing included testing notifications with BigPicture style on different Android devices without killing the app and with killing the app. 

- Tests covered:

        - Notification delay with the big picture with content title, summary text, and content description;
        - BigPicture instead of a large icon;
        - Big picture selected from the device gallery or from the resources and portrayed in the notification
        - Small icon and Large icon + URI;
        - Downloading and portraying image from the internet and portraying it in a Notification
 
The tested feature did not introduce any issues, everything works as expected, apart from some Android limitations that are mentioned in the testing document. More detailed testing results can be found in the [testing document ](https://docs.google.com/document/d/1SuH6ECDfsi-pwn7OAYwiypSGkFERkaKLV_FNxkmoHAk/edit?usp=sharing)

PR #239, opened 2022-12-08
Android permission fixes

aurimasc:
Removing uses of DeniedDontAskAgain, similar to Unity trunk. On Android 13 we don't reliably detect if it's a permanent dismiss, or a soft dismiss by swiping. And this is the only Android version where permission is required.

The changes after this release:
- AndroidNotificationCenter.UserPermissionToPost now should never return DeniedDontAskAgain
- PermissonRequest should work even if we have saved DeniedDontAskAgain is prefs

Testing required:
- Build master version using target SDK 33, request for permission and dismiss the dialog by swiping; you should get a permanent deny and not be able to request again
- Replacing the app with version from this PR should let to request again and again until Deny button is pressed twice, after which you can only go to settings and enable there
- the Main test project is just fine for this, has buttons to request permission and to open settings.

### DIFF
--- a/com.unity.mobile.notifications/CHANGELOG.md
+++ b/com.unity.mobile.notifications/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this package will be documented in this file.
 ## [Unreleased]
 
 ### Fixes:
+- [Android] - No longer use DeniedAndDontAskAgain permission response, only Denied.
 - [iOS] - Fix occasional crash when registering for push notifications.
 
 ## [2.1.0] - 2022-09-23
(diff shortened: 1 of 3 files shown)


todi1856 (approved):

guobyt (approved):
Tested the Main project with VLNQA00499 - Google Pixel 7 (Pixel 7), CPU: Google Tensor G2, GPU: Mali-G710, OS: 13

Tested:
1. Pressing the "Request permission" button and dismissing the dialog windows by swiping it multiple times
2. Pressing the "Request permission" button and pressing "don't allow" in the dialog window to disable notifications multiple times
3. Enabling notifications permissions from settings and checking if notifications arrive

Everything looks good. Dismissing the dialog window by swiping it doesn't permanently deny notifications permissions, after the second time denying notifications permission the dialog window doesn't open up, and after enabling notifications permission from the settings notifications arrive.

PR #241, opened 2022-12-13
Add support for Action icons

aurimasc:
https://jira.unity3d.com/browse/MNB-26

Since iOS 15 notification actions can have icons on them. This PR adds new APIs to support that.
There are two ways to specify an icon: system icon by name, or icon in app bundle by name (the later means an icon placed in XCode project).
In the test project the actionable notification button has been modified. The Reject action now has system icon on it, while Later has one from app bundle. Due to lack of API, the later doesn't work out of the box, there is an image file in the Main projects root directory that you have to drag into XCode to make it work.

I've tested it already, the only extra testing required is to see that older iOS versions don't blow up (no icon support there, but actions should work).


Pavel-Unity (approved):

guobyt (approved):
Tested added support for action icons works as expected, no new issues were found. Actions still work perfectly fine with iOS that are below the 15 version. The more detailed results are in the [testing document](https://docs.google.com/document/d/1Uv_OzljG58p5CeFvEDd_5c3rfVI5VfyRoyPGUJ2t6P8/edit?usp=sharing)

PR #247, opened 2023-01-09
Make logcat capture command always succeed

aurimasc:
In rare case test runs fail because command for capturing logcat fails or gets stuck for some reason. This is bad, because logcat is only captured to make it easier to debug things when something goes wrong, it is not essential. It is particularly bad when everything else is green, as in such case there is no use for logcat.
Modifying the command to always exit with success code, so that we don't have false-positives.


todi1856 (approved):
questionable change, maybe it would have been enough to clear the logcat before starting tests?

aurimasc:
> questionable change, maybe it would have been enough to clear the logcat before starting tests?

the last failure I encountered was failure to connect to deamon.

todi1856:
> > questionable change, maybe it would have been enough to clear the logcat before starting tests?
> 
> the last failure I encountered was failure to connect to deamon.

Maybe makes sense to add comment next to that line regarding deamon error, just to have an inline explanation why exit 0 is needed.

Issue #231, opened 2022-10-26
Mobile push notifications 2.1.0 not working on samsung devices running android 12


EoghanGallagher:
Hi . 

 I have a unity app that uses  mobile push notifications 2.1.0 . 

I have the correct permissions set for android devices running android 12  ( SCHEDULE_EXACT_ALARM ) .  Actually I have the schedule_at_exact_time drop down set to everything .

The notifications run perfectly on a moto g62 and a google pixel 4 running android 12 . 

However on a  samsung note 10+ running android  the notifications are inconsistent or don't fire at all .  This occurs on multiple 
samsung devices

Am i missing something!!

aurimasc:
Try looking into settings and what permissions are granted for your app. The permission to schedule exact alarm does not guarantee your app will be able to, that still can be disabled in settings.

Mirkobil:
same here Mobile Notification 2.1.0, "Schedule at exact time" set to "Everthing" works on all phones except samsung android 12.. On samsung galaxy s10 android 11 everything works fine, after updating to 12 everything stopped.. on samsung s22 ultra notifications work partially.. some fire some don't.. app notification permissions are fully granted on settings panel

EoghanGallagher:
@Mirkobil So I had a look at the settings on Samsung note 10 Plus.
  
I went to App -> my app -> Battery and then selected Unrestricted.
I checked ADB and noted that my app is now on the doze white list . 

adb shell dumpsys deviceidle

Selecting the other options Optimized and Restricted removed the app from the whitelist.

Notifications work fine even in battery saver mode when Unrestricted is selected

I encountered a new issue though on the Samsung galaxy a13  . 

I scheduled some test notifications. Then I enabled battery saver mode with all options selected. Turned it off again . The notifications never fired even with battery saver mode turned off.

EoghanGallagher:
What's the best approach for getting you app white listed ? 
 
Should I implement it myself ? I.E request REQUEST_IGNORE_BATTERY_OPTIMIZATIONS permissons an then attempt to add my app to the whitelist via code . Present the user with a prompt to allow the app to run in the background or grant unrestricted access to the battery?

xiabob:
maybe "[App hibernation](https://developer.android.com/topic/performance/app-hibernation)" issue?

leehans:
[...]

PR #249, opened 2023-01-24
Expansion of Android permissions for notifications

aurimasc:
Addresses: https://github.com/Unity-Technologies/com.unity.mobile.notifications/issues/231

Sometimes scheduled notifications do not arrive or are late, may even not arrive at all until you open the app. Only was able to reproduce it with battery saving on.

Changes:
- new flag for exact scheduling to add battery ignore permission to the manifest
- new public APIs to query if using exact scheduling and if ignoring batter optimizations
- new API to request exact scheduling (required on Android 12+ with SCHEDULE_EXACT permission)
- new API to request app being excluded from battery saving
- new API for notification permission rationale

Testing guide:
- find devices to reproduce the issue on; likely you need Android 12 or later and battery to be put to one of more restrictive power saving options; make sure you disable exact scheduling in notification settings, best to schedule notification couple minutes in the future, kill app and lock device, best if notification does not show up at all until you launch app again; some devices need to be unplugged from computer to be truly in power saving mode;
- there are two permissions for exact scheduling: SCHEDULE_EXACT_ALARM and USE_EXACT_ALARM. The later is Android 13 and should be enough for notifications to show up. The first one is since Android 12 and it allows app to request it, but you have control over it
- requesting exact scheduling or batter ignore requires relevant permission, set the corresponding flag in notification settings
- try requesting either exact scheduling or batter exempt (only one a the time, uninstall the app between tests or make sure settings are correct), I expect either to make notification arrive (but I'm not entirely sure, Samsung devices seem to be the most problematic here)
- there are 3 new buttons in Main test project: two for making relevant requests and one for querying both;
- test project on each focus gain prints notification permission status; only relevant on Android 13; if no permission given, it will print whether a UI with permission rationale should be shows (this is typically false initially and becomes true after permission was denied once).

Tested on:
- Asus ROG (8.1)
- Nokia 7 Plus (10)
- Galaxy S22 (12)
- Pixel 5 (13)
S22 was best for testing, as there with battery saving on inexact notifications didn't arrive at all until app is brought to foreground (at least withing reasonable amount time, didn't wait for hours).

### DIFF
--- a/TestProjects/Main/Assets/Scripts/AndroidTest.cs
+++ b/TestProjects/Main/Assets/Scripts/AndroidTest.cs
@@ -328,6 +328,10 @@ namespace Unity.Notifications.Tests.Sample
         {
             var isExact = AndroidNotificationCenter.UsingExactScheduling;
             m_LOGGER.Blue($"Scheduling at exact time: {isExact}");
+        }
+
+        void RequestExactScheduling()
+        {
             AndroidNotificationCenter.RequestExactScheduling();
         }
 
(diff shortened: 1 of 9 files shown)


todi1856 (approved):

guobyt (approved):
Tested expansion of Android permissions for notifications. It includes testing of USE_EXACT_ALARM, SCHEDULE_EXACT_ALARM and IGNORE_BATTERY_OPTIMIZATIONS permissions. Notifications arrive at the exact time but some are late - the battery ignore setting helps to fix this problem. Also, these exceptions are limited by the device. One issue not introduced by the PR was found and is now being fixed - [MNB-36](https://jira.unity3d.com/browse/MNB-36). Overall everything looks great. The more detailed results are in the [testing document](https://docs.google.com/document/d/1nAU3LARCmuTSO3nnKydErYOxqkmlnznx2gLe_wCs-us/edit?usp=sharing)

PR #251, opened 2023-01-30
iOS notification content update

aurimasc:
https://jira.unity3d.com/browse/MNB-27
Adds two new properties to iOSNotification: InterruptionLevel and RelevanceScore.

Both of these have been introduced in iOS 15, on lower versions they should be ignored.
InterruptionLevel determines if/when notification causes alert (see Apple docs).
RelevanceScore works when sending multiple notifications and collapsing them shows the one with highest score. Apple docs are very non-descriptive for this, I've tried using the same thread id, there's also category that I haven't tried.

Testing required:
- no problems on iOS <15
- all interruption levels work (test project has added button for  low importance)
- check the relevance too, note that when sending multiple notifications with duplicate score it did show unexpectedly for me, but with one notification standing out in score it looked as expected.

### DIFF
--- a/TestProjects/Main/Assets/Resources/iOSNotifications/TimeIntervalTrigger/Passive.asset
+++ b/TestProjects/Main/Assets/Resources/iOSNotifications/TimeIntervalTrigger/Passive.asset
@@ -1,0 +1,31 @@
+%YAML 1.1
+%TAG !u! tag:unity3d.com,2011:
+--- !u!114 &11400000
+MonoBehaviour:
+  m_ObjectHideFlags: 0
+  m_CorrespondingSourceObject: {fileID: 0}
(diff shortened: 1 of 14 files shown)


Pavel-Unity (approved):

robertas-unity (approved):
Testing done - looks good to me.

RelevanceScore works - notifications are shown correctly in the notification summary, with the highest relevance score showing up on top.
Different InterruptionLevels work correctly as well.

Tested with iOS 15 & iOS 16 devices.
Devices with iOS < 15 still work correctly even with the properties being used, they are simply ignored.

Full testing details: https://docs.google.com/document/d/1CovESxZCUmqp3h3Oj9IQOwOqC2LRZPuoM9A1CcfzCrk/edit?usp=sharing

PR #253, opened 2023-02-01
Android api level cleanup

aurimasc:
Removes hardcoded numbers and use of reflection from Java code. Also specifies the required API level is documentation.
Unity infrastructure has been updated and API level 31 is now available for all versions. It is also the level new apps must target to be accepted to Google Play, so it looks reasonable to require such SDK when building.

No user visible changes other than API level requirement in this PR. I don't think there is a reason to test anything here (of course it won't hurt to do it).


todi1856 (approved):
Do you need to add entries in changelog?

aurimasc:
> Do you need to add entries in changelog?

There are no user visible changes here. A micro optimization of not using reflection.

guobyt (approved):
Looks good to me, no test is needed since these changes are not visible to the users.

PR #255, opened 2023-02-23
Formatting update

aurimasc:
Two changes here:
- Check format for the whole package
- Use specific version of formatter

Main change is the later. It happened more than once when formatter update actually did minor changes to format and perfectly formatted code became incorrectly formatted. This shouldn't happen, repository should use a specific fixed formatter and consciously update it.

### DIFF
--- a/.yamato/format.yml
+++ b/.yamato/format.yml
@@ -5,7 +5,8 @@ check_format:
     image: package-ci/ubuntu:stable
     flavor: b1.small
   commands:
-    - git clone --branch stable git@github.cds.internal.unity3d.com:unity/unity-meta.git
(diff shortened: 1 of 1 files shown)


todi1856 (approved):
Not sure I agree, because that means your package might not follow latest official formatter rules. So it's possible other Unity package will have different formatting, than your package.

Sure they failed with stable formatter, but that was a mistake, not sure if you should try workarounding their mistake.

In any case, this is up to you.

aurimasc:
> Not sure I agree, because that means your package might not follow latest official formatter rules.

That's the problem: the formatter defines what the rules are, not merely enforce them. And new versions of formatter come out with not much warning. A well formatted code should not become badly formatted without code changes, that's why I'm pinning the version.

PR #257, opened 2023-02-28
GameActivity support

aurimasc:
Support GameActivity. Our notifications open an activity when tapped.
Changes:
- Do not use current activity as an option anymore; this is wrong - it only works as long as you don't kill the app, while for notifications the most essential part is precisely to work well after the app was killed.
- Fail initialization if we cannot determine the activity to open, so user knows in advance that something is wrong.
- If activity to be opened is not explicitly specified by the user, loop over activities in the application and select either the Unity one or the only one, fail if there are multiple choices.

Tested by developer/automation:
- Having either Activity or GameActivity chosen works as expected
- Having both enabled in Player Settings fails initialization logging the exact reason

Testing required:
- Explicitly specify the activity class in notification settings, that one should be used no matter how many other activities you have
- Having extra activities in project should pick the Unity activity, if only one such is present
- When project has only one enabled activity, that one should be used
- Try reschedule after boot
- Hints:
  - Not specific to Android device or version, Unity version only matters for GameActivity part
  - The easiest way to test is to enable both activity types and specifying custom activity for notifications, then export the Android Studio project and play there by modifying manifest in the unityLibrary project and adding activities via IDE (custom activity is in the manifest as meta data)
  - You can enable/disable activities in the manifest, we only care about enabled ones
  - To have no Unity activity, make a class that extends one of Unity activities and use this new class in the manifest instead of Unity one

### DIFF
--- a/com.unity.mobile.notifications/Runtime/Android/Plugins/com/unity/androidnotifications/UnityNotificationUtilities.java
+++ b/com.unity.mobile.notifications/Runtime/Android/Plugins/com/unity/androidnotifications/UnityNotificationUtilities.java
@@ -510,7 +510,7 @@ class UnityNotificationUtilities {
         return null;
     }
 
-    protected static Class<?> getOpenAppActivity(Context context, boolean fallbackToDefault) {
+    protected static Class<?> getOpenAppActivity(Context context) {
         ApplicationInfo ai = null;
         try {
             ai = context.getPackageManager().getApplicationInfo(context.getPackageName(), PackageManager.GET_META_DATA);
@@ -532,7 +532,7 @@ class UnityNotificationUtilities {
             }
         }
 
-        if (activityClass == null && fallbackToDefault) {
+        if (activityClass == null) {
             Log.w(TAG_UNITY, "No custom_notification_android_activity found, attempting to find app activity class");
 
             String classToFind = "com.unity3d.player.UnityPlayerActivity";
--- a/com.unity.mobile.notifications/Runtime/Android/Plugins/com/unity/androidnotifications/UnityNotificationUtilities.java
+++ b/com.unity.mobile.notifications/Runtime/Android/Plugins/com/unity/androidnotifications/UnityNotificationUtilities.java
@@ -519,14 +519,11 @@ class UnityNotificationUtilities {
         }
         Bundle bundle = ai.metaData;
 
-        String customActivityClassName = null;
         Class activityClass = null;
 
         if (bundle.containsKey("custom_notification_android_activity")) {
-            customActivityClassName = bundle.getString("custom_notification_android_activity");
-
             try {
-                activityClass = Class.forName(customActivityClassName);
+                return Class.forName(bundle.getString("custom_notification_android_activity"));
             } catch (ClassNotFoundException ignored) {
                 ;
             }
--- a/com.unity.mobile.notifications/Runtime/Android/Plugins/com/unity/androidnotifications/UnityNotificationUtilities.java
+++ b/com.unity.mobile.notifications/Runtime/Android/Plugins/com/unity/androidnotifications/UnityNotificationUtilities.java
@@ -11,6 +11,7 @@ import android.app.Notification;
 import android.content.Context;
 import android.content.Intent;
 import android.content.SharedPreferences;
+import android.content.pm.ActivityInfo;
 import android.content.pm.ApplicationInfo;
 import android.content.pm.PackageManager;
 import android.content.res.Resources;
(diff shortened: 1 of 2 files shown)


todi1856 (changes requested):

com.unity.mobile.notifications/Runtime/Android/Plugins/com/unity/androidnotifications/UnityNotificationUtilities.java
```diff
@@ -510,46 +511,72 @@ private static <T extends Parcelable> T deserializeParcelable(DataInputStream in
...
-                } catch (ClassNotFoundException ignored1) {
-                    Log.w(TAG_UNITY, String.format("Attempting to find class based on package name: %s, failed!", classToFind));
+            ActivityInfo[] aInfo = pm.getPackageInfo(context.getPackageName(), PackageManager.GET_ACTIVITIES).activities;
+            if (aInfo != null) {
+                String activityClassName = null;
+                String conflictingActivity = null;
+                boolean activityConflict = false;
+                for (ActivityInfo info : aInfo) {
+                    // activity alias not supported
+                    if (info.enabled && info.targetActivity == null) {
```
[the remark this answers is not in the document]
aurimasc: Won't that only spread thread more vertically and still have the same problem?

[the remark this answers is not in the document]
todi1856: I was think something like this, this will also makes easier code to expand in case we'll have more activities in the future:

```
            ActivityInfo[] aInfo = pm.getPackageInfo(context.getPackageName(), PackageManager.GET_ACTIVITIES).activities;
            if (aInfo == null)
                return null;

            List<String> validActivities = new ArrayList<String>();
            for (ActivityInfo info : aInfo) {
                if (!info.enabled || info.targetActivity != null)
                    continue;

                if (info.name.endsWith(".UnityPlayerActivity") ||
                    info.name.endsWith(".UnityPlayerGameActivity"))
                    validActivities.add(info.name);
            }

            if (validActivities.size() > 0)
            {
                Log.e(TAG_UNITY, String.format("Multiple choices for activity for notifications: %s", String.join(", ", validActivities)));
                return null;
            } else if (validActivities.size() == 0)
            {
                Log.e(TAG_UNITY, "Activity class for notifications not found");
                return null;
            }
            else
            {
                return Class.forName(validActivities.get(0));
            }

        } catch (PackageManager.NameNotFoundException e) {
            e.printStackTrace();
        } catch (ClassNotFoundException e) {
            Log.e(TAG_UNITY, "Failed to find activity class: " + e.getMessage());
        }

        return null;
```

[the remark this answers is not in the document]
aurimasc: This is also not correct :)
Any activity is fine as long as there is no .UnityPlayer*Activity. Will adding that extra logic still make it smaller and clearer?
Note, that this code runs when app is in background, so avoiding extra allocations and extra time spent is desirable. Not super crucial, but it should be a significant benefit in code clarity.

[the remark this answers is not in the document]
todi1856: Not sure if extra allocations is a problem here, but you current code is barely readable, please refactor :)

com.unity.mobile.notifications/Runtime/Android/Plugins/com/unity/androidnotifications/UnityNotificationUtilities.java
```diff
@@ -533,43 +533,47 @@ protected static Class<?> getOpenAppActivity(Context context) {
...
+                    activityClassName = info.name;
+                    activityConflict = false;
+                } else {
+                    conflictingActivity = info.name;
+                    activityConflict = true;
                 }
+            }
+
+            if (activityConflict) {
+                Log.e(TAG_UNITY, String.format("Multiple choices for activity for notifications: %s and %s", activityClassName, conflictingActivity));
```
todi1856, line 567: if there are two conflicting activities, this error will only print one, shouldn't it print all conflicting activities ?

todi1856: what settings?

aurimasc: Notification settings. At the start of this method we take the activity from the manifest (put there during build) and this loop is a fallback. If you have multiple candidate activities, then you have to specify one in settings.

todi1856: That's fine and all, but I am talking about the error message being incomplete, it won't mention to you all activities?

aurimasc: It's pointless to print them all out.
But maybe it's worth to change the message to:
"Multiple choices for activity, please explicitly specify fully qualified activity class in Notification Settings"

todi1856: generic message would be better, because with current implementation, it's not clear how you decide what activities to print in error message

todi1856 (approved):
Leaving up to you to decide how to print error message

guobyt (approved):
Tested added GameActivity support. Additionally, tested Custom Activities, rescheduling after the device restarted, killing the app, and putting it in the background. No new issues were found, and everything works as expected. The more detailed results are in the [testing document](https://docs.google.com/document/d/1Hjv5inje00dXpdFn82rSoga1fd6eDEWKQGeNamJDF-w/edit?usp=sharing)

PR #259, opened 2023-03-07
Android proguard seetings

aurimasc:
https://jira.unity3d.com/browse/MNB-38
Add proguard settings for notifications during post process. Otherwise if you enabled minification in player settings, notifications will not work because Java code will be obfuscated. The convention currently is to make stuff used from C# public, some constants are made such as a result.

To test: enable minification in Player Settings publishing section and try both release and development builds. Take a random sample of actions in Main test project. All should work as expected. Not specific to Unity version nor device nor Android version, but as two runs are required, probably good idea to do them on different Unity versions.

### DIFF
--- a/com.unity.mobile.notifications/Editor/AndroidNotificationPostProcessor.cs
+++ b/com.unity.mobile.notifications/Editor/AndroidNotificationPostProcessor.cs
@@ -273,21 +273,29 @@ namespace Unity.Notifications
             }
 
             var lines = File.ReadAllLines(proguardFile);
-            string[] newLines;
-            if (InjectProguard(lines, out newLines))
-                File.WriteAllLines(proguardFile, newLines);
+            if (InjectProguard(ref lines))
+                File.WriteAllLines(proguardFile, lines);
         }
 
-        internal static bool InjectProguard(string[] original, out string[] result)
+        internal static bool InjectProguard(ref string[] lines)
         {
-            foreach (var s in original)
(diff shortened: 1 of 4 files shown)


com.unity.mobile.notifications/Editor/AndroidNotificationPostProcessor.cs
```diff
@@ -260,6 +262,34 @@ internal static void AppendAndroidMetadataField(string manifestPath, XmlDocument
...
+            var lines = File.ReadAllLines(proguardFile);
+            string[] newLines;
+            if (InjectProguard(lines, out newLines))
+                File.WriteAllLines(proguardFile, newLines);
+        }
+
+        internal static bool InjectProguard(string[] original, out string[] result)
+        {
+            foreach (var s in original)
+                if (s.Contains("com.unity.androidnotifications.UnityNotificationManager"))
```
todi1856, line 284: wouldn't it be more correct to search for presence of the string you're adding it, meaning
```
"-keep class com.unity.androidnotifications.UnityNotificationManager { public *; }"
```

aurimasc: An existing project may already have this setup, so I'd rather search for class.

todi1856: Alright, hopefully nobody will add an entry like com.unity.androidnotifications.UnityNotificationManagerMyEx

todi1856 (approved):

PR #261, opened 2023-03-17
Android notification update fix

aurimasc:
https://jira.unity3d.com/browse/MNB-41
Fix notification update for the case where app is not killed completely.
Also a minor fix where notification update was regarded as user customized, when it actually wasn't and was using more expensive serialization as a result.

There are two manual tests for this feature in Main test project:
- schedule notification with explicit id and press update button before it arrives -> you should get a notification with text telling that it was updated (without clicking the update button notification should show it's id)
- When updated notification is present in taskbar, scheduling the explicit id one again should replace the existing notification in taskbar.
The changes are not specific to Android version nor Unity version nor device.

### DIFF
--- a/com.unity.mobile.notifications/CHANGELOG.md
+++ b/com.unity.mobile.notifications/CHANGELOG.md
@@ -14,6 +14,9 @@ All notable changes to this package will be documented in this file.
 - [iOS] - Added property to set interruption level (requires iOS 15).
 - [iOS] - Added property to set relevance score (requires iOS 15).
 
+### Fixes:
+- [Android] - Fix updating scheduled notification not updating if app is not killed.
+
 ## [2.1.1] - 2023-01-04
 
 ### Fixes:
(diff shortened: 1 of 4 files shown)


todi1856 (approved):

guobyt (approved):
The testing involved scheduled notifications with a specific ID and updating them before they were due to be delivered. When an updated notification arrived then a notification with the explicit ID was sent once again to replace the existing one only with an ID. The testing was carried out by both keeping the app open in the foreground and killing it, and by testing both Application Entry Points - Activity and GameActivity. Everything looks good.

**Tested with devices:** 
- VLNQA00460 - Google Pixel 6 (Pixel 6), CPU: Google Tensor (Whitechapel), GPU: Mali-G78, OS: 13
- VLNQA00467 - Galaxy S21 Ultra 5G (SM-G998B), CPU: Exynos 2100, GPU: Mali-G78, OS: 12
- VLNQA00230 - OnePlus 6T (ONEPLUS A6013), CPU: Snapdragon 845 SDM845, GPU: Adreno 630, OS: 9.0.0

PR #264, opened 2023-03-21
Android: report if notifications are enabled

aurimasc:
https://jira.unity3d.com/browse/MNB-43
On Android 7+ notifications can be disabled per app. Right now we have AndroidNotificationCenter.UserPermissionToPost, but it only report if permission is enabled, which is Android 13+.
This PR adds one more enum value for disabled notifications and reports that on Android 7+. On even older Android versions it should be always enabled.
Note: looks like there is no such thing as disabled notifications in Android 13, it simply revokes the permission.

The Main test project prints permission status each time when gaining focus and there is a button to open settings. Please do try Android version <7 if nothing bad happens!

### DIFF
--- a/com.unity.mobile.notifications/CHANGELOG.md
+++ b/com.unity.mobile.notifications/CHANGELOG.md
@@ -11,6 +11,7 @@ All notable changes to this package will be documented in this file.
 - [Android] - Added API to check if UI with permission rationale for notifications should be shown before requesting permission.
 - [Android] - Added APIs to request exact scheduling and bupasssing battery optimizations.
 - [Android] - Minification seetings now managed automatically by package.
+- [Android] - AndroidNotificationCenter.UserPermissionToPost now will also report if notifications are blocked in Settings (Android 7+).
 - [iOS] - Added support for notification action icons (requires iOS 15).
 - [iOS] - Added property to set interruption level (requires iOS 15).
 - [iOS] - Added property to set relevance score (requires iOS 15).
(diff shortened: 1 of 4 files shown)


todi1856 (approved):

guobyt (approved):
The testing involved verifying when notifications were disabled in the app settings and if the logs reflected this change after the app was reopened from the background or after it was killed. The tests also included checking whether the device receives notifications after the app was killed, put in the background, kept in the foreground, or notification is rescheduled after a device restart.

With Android 5.0.1 device more detailed tests were done, to check if notifications are working as expected. It includes testing empty, with body, large icon, small icon, both icon, explicit ID notifications, modifying explicit ID notification and reschedule after restart.

No new issues were found everything works as expected.

Tested with devices:

- VLNQA00467 - Galaxy S21 Ultra 5G (SM-G998B), CPU: Exynos 2100, GPU: Mali-G78, OS: 12
- VLNQA00119 - Xiaomi Mi 5 (MI 5), CPU: Snapdragon 820 MSM8996, GPU: Adreno 530, OS: 8.0.0
- VLNQA00230 - OnePlus 6T (ONEPLUS A6013), CPU: Snapdragon 845 SDM845, GPU: Adreno 630, OS: 9.0.0
- VLNQA00090 - Samsung Galaxy S8 (USA) (SM-G950U), CPU: Snapdragon 835 MSM8998, GPU: Adreno 540, OS: 7.0.0
- VLNQA00140 - Samsung Galaxy S4 (GT-I9505), CPU: Snapdragon S4 Plus MSM8960, GPU: Adreno 320, OS: 5.0.1

Issue #271, opened 2023-05-26
ANR: Broadcast of Intent


ngolagani:
We are on package version 2.1.1 and Unity version 2020.3.45f1, seeing the following ANRs related to Unity Notifications. These are the different kinds of ANRs seeing on our game.

```
unavailable - com.unity.androidnotifications.UnityNotificationUtilities.deserializeNotification
Broadcast of Intent { cmp=com.ea.tsb/com.unity.androidnotifications.UnityNotificationManager }
```

```
Native method - android.os.MessageQueue.nativePollOnce
Broadcast of Intent { cmp=com.ea.tsb/com.unity.androidnotifications.UnityNotificationManager }
```

```
android.app.Instrumentation.callApplicationOnCreate
Broadcast of Intent { cmp=com.ea.tsb/com.unity.androidnotifications.UnityNotificationManager }
```

```
Native method - java.lang.Thread.sleep
Broadcast of Intent { cmp=com.ea.tsb/com.unity.androidnotifications.UnityNotificationManager }
```

```
Native method - android.os.MessageQueue.nativePollOnce
Broadcast of Intent { act=android.intent.action.BOOT_COMPLETED cmp=com.ea.tsb/com.unity.androidnotifications.UnityNotificationRestartOnBootReceiver }
```

```
Native method - jdk.internal.misc.Unsafe.park
Broadcast of Intent { cmp=com.ea.tsb/com.unity.androidnotifications.UnityNotificationManager }
```

ngolagani:
We are seeing it across all kind of devices and OS versions. And we don't have specific reproducible steps but looks like happening when we leave device ideal for some time.

aurimasc:
Do you have more complete stack traces?
In particular, the UnityNotificationUtilities.deserializeNotification seems to be the one that is causing the issue.

mariusJanke:
We are also seeing the issue on a project. 
We are using Unity 2021.3.21f1, with the Package Version 2.1.1

Here is the full stack trace for the UnityNotificationUtilities.deserializeNotification:
```
at java.lang.Object.wait (Native method)
  at java.lang.Object.wait (Object.java:442)
  at java.lang.Object.wait (Object.java:568)
  at android.app.SharedPreferencesImpl.awaitLoadedLocked (SharedPreferencesImpl.java:279)
  at android.app.SharedPreferencesImpl.getString (SharedPreferencesImpl.java:301)
[...]

PR #278, opened 2023-08-16
Deserialize on background thread

aurimasc:
https://github.com/Unity-Technologies/com.unity.mobile.notifications/issues/271
It is not advised to do IO on Android UI thread, as that may take time and lead to ANRs. Solution is to offload the work onto a background thread.
There are two places where the issue is present:
- receiving a broadcast that notification needs to be shown
- receiving broadcast that device has been rebooted and notifications have to be rescheduled
Two changes not related to the bug:
- Logging when reboot message is received, so it can be debugged whether it's app issue, or Android did not send app the message (which does happen on some devices)
- Give a 10 minute slack when rescheduling after boot. With Android pushing towards inexact alarms, notifications showing a couple minutes late is becoming a norm, so rescheduling should to some degree match. 10min is chosen arbitrary, we don't want to show notifications that expired a long time ago.

Developer testing:
- Unity trunk only (should not matter aside that trunk only supports Android 6, while older versions support 5.1)
- Main test project from the package repo
- Scenarios:
  - Notifications scheduled to arrive after some time do arrive with app in foreground, background or completely killed
  - Scheduled after reboot notification (dedicated button in project) arrives after device reboot
- Devices tested with no issues:
  - Nokia 7 Plus (10.0)
  - Pixel 5 (13)
  - Galaxy S22 (12)
  - OnePlus6T (9.0)
  - Asus ROG Phone (8.1)

Testing request:
- Android only
- Try same scenarios on more devices, especially lower Android versions
- When testing reboot, capture logcat for all processes and search for "resched", if the app doesn't print anything, it's a device issue (launching app quickly may be a solution); it is known that some devices don't work well in this scenario.
- Some devices have auto-start feature. App may be have auto-start disabled for it by default and experience issues because of that. Assu ROG had this and I would not get notifications after reboot of if app is killed. Enabling auto-start for app resolved the problem.

### DIFF
--- a/com.unity.mobile.notifications/CHANGELOG.md
+++ b/com.unity.mobile.notifications/CHANGELOG.md
@@ -5,8 +5,10 @@ All notable changes to this package will be documented in this file.
 ## [Unreleased]
 
 ### Changes & Improvements:
+- [Android] - Reschedule after reboot will send all notifications that expired less than 10 minutes ago.
 
 ### Fixes:
+- [Android] - [issue 271](https://github.com/Unity-Technologies/com.unity.mobile.notifications/issues/271) Fix possible ANR when sending/rescheduling after reboot.
 
 ## [2.2.1] - 2023-07-17
 
--- a/com.unity.mobile.notifications/Runtime/Android/Plugins/com/unity/androidnotifications/UnityNotificationManager.java
+++ b/com.unity.mobile.notifications/Runtime/Android/Plugins/com/unity/androidnotifications/UnityNotificationManager.java
@@ -719,36 +719,39 @@ public class UnityNotificationManager extends BroadcastReceiver {
 
         Notification notif = null;
         int id = -1;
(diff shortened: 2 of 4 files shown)


todi1856 (approved):

IGuscin (approved):
Tested on couple more devices.
Notifications works as expected on these devices (notifications are arriving when app in foreground, background, killed. Scheduled notifications also arrive after reboot): 
LG V30 Android 8.0.0
Galaxy Z Fold2 5G Android 11
Motorola Nexus 6 Android 7.1.1
Pixel 6 PRO Android 13
Xiaomi Redmi Note 9 Pro Android 12
These devices are failing to show notifications when app is killed and after reboot:
Xiaomi Redmi 6 Pro Android 9.0.0
Xiaomi Mi 5s Android 6.0.1

Seems to be widely known issue, related to Autostart/battery saving settings, not related to this PR or Unity notifications package at all

PR #280, opened 2023-08-24
iOS remote notification presentation fixes

aurimasc:
https://jira.unity3d.com/browse/MNB-60
https://jira.unity3d.com/browse/MNB-61

Fix two bugs related to iOS remote notification presentation:
- Remote presentation options set in settings are only used if permission is requested on app startup, not if manually requested by user at later point, notification does not show up as a result when delivered to foreground app
- If callback for remote notification receive is registered, notification is not shown when app is in the foreground

One small feature added while fixing this: adding showInForeground key to remote notification now allows to show it when received with app in foreground (just like for local notifications).

Both fixes verified by developer:
- requesting permission at later point in the app still shows remote notification sent to foreground app
- registering callback for remote notification does not alter the behavior except that now notification is shown based on showInForeground flag on it (without callback it is always shown).

### DIFF
--- a/com.unity.mobile.notifications/CHANGELOG.md
+++ b/com.unity.mobile.notifications/CHANGELOG.md
@@ -6,9 +6,12 @@ All notable changes to this package will be documented in this file.
 
 ### Changes & Improvements:
 - [Android] - Reschedule after reboot will send all notifications that expired less than 10 minutes ago.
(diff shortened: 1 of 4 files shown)


PetrasUnity (approved):
LGTM

guobyt (approved):
Sufficiently tested by the developer, no additional QA pass needed.

PR #285, opened 2023-09-04
Refactor Android to use .androidlib

aurimasc:
https://jira.unity3d.com/browse/PLAT-6611
Modify Android setup to use .androidlib instead of loose .java files. There are multiple benefits from this:
- The Java code that depends on notifications Java code now must also be loose .java files, which causes problems with external dependencies; with .androidlib depending code can also be .androidlib with proper dependencies (push notification package encountered this)
- Being gradle module allows having a manifest with always required receivers and permissions, avoiding file patching in the unityLibrary
- Gradle module can have proguard rules files, no need to patch unityLibrary
- Gradle module can specify compile sdk; Unity specifies it to the same as target SDK, which makes notifications code not compile if set too low

Testing:
- no functional changes to package, only risks of project related regressions
- tried building with target SDK 29 (Unity 2021.3) - works fine as long as SDK 33 is installed, gives reasonably clear error without SDK 33 (gradle error, but you can see sdk 33 being a problem)
- release build with minification enabled works (testing proguard rules)
- we are now have separate manifest file and are patching it instead of the main one from unityLibrary; did not try all possible features, but restart after reboot works and optional permissions are present in modules manifest as well as merged manifest of the apk, so assuming everything else works too
- icons are now placed into notification module in gradle project instead of unityLibrary; notification with large icon works and displays correctly, so icon files correctly end up in the apk


todi1856 (approved):
Nice

com.unity.mobile.notifications/Runtime/Android/Plugins/mobilenotifications.androidlib/build.gradle
```diff
@@ -0,0 +1,27 @@
+plugins {
+    id 'com.android.library'
+}
+
+android {
```
IGuscin, line 5: buildToolsVersion '30.0.2' is specified in build.gradle of unityLibrary if built with Unity 2021.3 targeting SDK 33 (SDK is downloaded on request). buildToolsVersion is not specified for MobileNotifications and by default is requiring 29.0.2, but fails to install. Maybe we should match build tools version used for unityLibrary?

IGuscin (changes requested):
Overall looks good, checked that manifests are merged correctly, minification doesn't break anything. Notifications woks as expected.
Few concerns from me:
1. If SDK 33 is not present - current workflow is not user friendly. Build fails with gradle error. Not very easy to find the reason and a way to resolve that.
2. Targeting SDK 33 prompts to update SDK, downloads Android 33 platform. But then build fails with error that build tools 29.0.2 failed to install. Not sure if we have to stick to 29.0.2, unityLibrary build.gradle specifies 30.0.2 which is already present

com.unity.mobile.notifications/Runtime/Android/Plugins/mobilenotifications.androidlib/build.gradle
aurimasc: It seems that build tools are not fixed, Unity uses what it finds in SDK.
I've installed latest 22.3 and build tools there are 32.0.0, looks to be the same in all later releases.
Worst part is, we don't expose it, so there is no way of knowing.

IGuscin (approved):

IGuscin:
Now works a bit better, thanks. Tested that with latest changes mobileNotifications uses the same buildTools asunityLibrary and compileSDK either 33 (if unityLibrary uses lower version) or the same as unityLibrary. 
Initial problem, when user is missing SDK 33 and targets lower version still exists, but current approach is much better and user friendly. Also, noticed not PR related issue: Editor doesn't suggest to install missing SDK if a higher version is installed. For example install SDK 34 and then try targeting SDK 33 (not installed yet). Build will just fail with missing SDK error

todi1856 (approved):

Issue #277, opened 2023-08-07
[iOS] unregisterForRemoteNotifications ?


alexandrefresnais:
Hello guys,

Thanks for the great work, I managed to implement notifications in my iOS app quickly.

I am wondering why I see no possible call to `unregisterForRemoteNotifications` for push notifications ?

It seems to be the right way to unsubscribe from notifications when a user logs out.
For sure I can discard the token on my backend but calling `unregisterForRemoteNotifications` would enforce it locally even without internet connection.

Are you planning to add this method ?

Thank you in advance for reading me.

aurimasc:
Implemented.

PR #288, opened 2023-09-20
Unregister push support

aurimasc:
https://github.com/Unity-Technologies/com.unity.mobile.notifications/issues/277
iOS has APIs not only to register for push notifications, but also to unregister. While rarely used, it has use cases, such us user logging out of app.
This PR adds two new methods: to check if registered and to unregister.

Added two buttons to the Maintest project for each of new APIs:
- "Check push registration"
- "Unregister push notifications"
Tested both:
- the app registers for push on launch, so checking returns true
- after pressing unregister check reports push is not registered
- Pressing button to request authorization register for push again (check button returns true)
Tested on iPhone X iOS 15.3.1
Trying few more iOS versions would be nice, otherwise it's quite straight-forward.

### DIFF
--- a/TestProjects/Main/Assets/Scripts/iOSTest.cs
+++ b/TestProjects/Main/Assets/Scripts/iOSTest.cs
@@ -195,6 +195,11 @@ namespace Unity.Notifications.Tests.Sample
             {
                 StartCoroutine(GetCurrentLocation());
             });
+            m_groups["General"]["Check push registration"] = new Action(() =>
+            {
+                RegisteredForRemoteNotifications();
+            });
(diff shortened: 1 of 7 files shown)


AurochsCross (approved):

PetrasUnity (approved):

guobyt (approved):
Tested with:
- VLNQA00352 - iPhone 12 Pro (MGML3ET/A), CPU: t8101, GPU: Apple designed, OS: 16.2
- VLNQA00476 - iPhone 13 Pro Max (MLL63ET/A), CPU: t8110, OS: 17.0

With the Main project checked:
- if the app registers for Push Notification on launch
- If it's possible to unregister Push Notification
- If the app registers for Push notifications again after requesting authorization

Everything works as expected.

Issue #290, opened 2023-09-27
[Android] Large Icon is not displayed correctly


akomekagome:
I set LargeIcon in the Android notification settings but it is not displayed correctly. I investigated and it seems that Return was not done at the relevant location.

https://github.com/Unity-Technologies/com.unity.mobile.notifications/blob/de5741686b24e9056b83ae0a6bdc87d713543e63/com.unity.mobile.notifications/Runtime/Android/Plugins/com/unity/androidnotifications/UnityNotificationManager.java#L833

PR #292, opened 2023-09-29
Fix large icon from file

aurimasc:
https://github.com/Unity-Technologies/com.unity.mobile.notifications/issues/290
Large icon does not show up on notification when provided as file path. A fix is trivial and is very unlikely to cause any side effects.
A test project for big picture style update to add a case for this + project update with regard to build tools version.

Tested using mentioned project for big picture support. Issue is not specific to neither Unity nor Android version/device.
In my opinion no additional testing is required.


todi1856 (approved):

guobyt (approved):
Sufficiently tested by the developer, no additional QA pass needed.

PR #294, opened 2023-10-03
iOS presentation option update

aurimasc:
https://jira.unity3d.com/browse/PLAT-5493
Apple introduced new presentation options we don't have support for yet. This PR adds them.
The options apply when notification is in foreground. Meaning:
- List - notification will be present in notification list until tapped (swipe from the top to see the list)
- Banner - a visual notification banner will show up above the application (looks like the same as Alert which is now deprecated in Apple docs)

Modified few notifications in Main test project to use these new settings:
- "Empty notification + body (silent)" has banner, no list
- "Empty notification + title (list)" has banner and list

On iOS 16.3:
- having Banner, but not List - shows notification above the app, once it's gone, nothing present in the list
- having Banner + List - shows notification above the app and in the list later until tapping notification

Concerns:
- two new options are iOS 14+, we support older iOS (likely they do nothing there, but I don't have the device to confirm)


PetrasUnity (approved):

aurimasc:
Tested on iPhone SE iOS 13.4, no issues, only the new flags are ignored (expected).

robertas-unity (approved):
Sufficiently tested by the developer, no additional QA needed.

PR #298, opened 2023-11-24
Fix plugin settings for .androidlib

aurimasc:
The .androidlib is marked for all platforms now, so as a result it is copied into Xcode project when built. No harm, as it's ignored, but incorrect too.
Tested manually that the library project is not copied to iOS any more. Automation tests that it is correctly copied into Android build.
Opened a bug in Unity, as .androidlib should only be compatible with Android by default .

### DIFF
--- a/com.unity.mobile.notifications/CHANGELOG.md
+++ b/com.unity.mobile.notifications/CHANGELOG.md
@@ -2,7 +2,12 @@
 
 All notable changes to this package will be documented in this file.
 
+## [Unreleased]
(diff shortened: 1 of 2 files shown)


todi1856 (approved):

JacobSimpson-unity (approved):
Sufficiently tested by the developer, no additional QA pass needed. This is mentioned [.androidlib bug](https://jira.unity3d.com/browse/UUM-57229).

PR #299, opened 2023-11-29
iOS: schedule using local/UTC time natively

aurimasc:
https://jira.unity3d.com/browse/MNB-66
We used to schedule notifications using UTC time when  using calendar trigger. This causes problems with daylight saving time. This PR changes to use either UTC or local time natively in Objective-C depending on how calendar trigger is setup.
Note, that users report has an incorrect expectation in switching timezones. iOS does schedule notification using current timezone, so switching it does not alter the time notification will arrive.

Tested on iPhone X, iOS 15.3 (though, the issue is not device or OS specific).
Used users project, but modified it. It is better to test that notification arrives, rather than the opposite. The main test scenario is: schedule notification to happen 1 hour and 2 minutes in the future, then change the phone time. This gives 6 test cases:
1) change phone time to one hour forward, notification arrives within two minutes
2) have phone time one hour behind and after scheduling notification switch back to automatic time, notification arrives withing couple minutes
3, 4) same as the two above except scheduling using UTC time and switching timezone after switching, then changing time; timezone change should have no impact, changing phone time one hour forward makes notification arrive within couple minutes
5, 6) same as first two except timezone on phone is changed after scheduling, then phone time is changed to one hour forward and notification should arrive within couple minutes

Note, that when notification is scheduled, it is printed in the log after how many seconds ti should arrive (telling you how much you should change the clock time).

Manually tested all scenarios with changing phone time manually. With switching back to automatic only tried local time, assuming other two will work as well.
Calendar trigger can be setup to fire when particular components match, one could use for example day, hour and minute only with repeating, then notification should happen each month when the day, hour and minute matches. Did not try this scenario.

### DIFF
--- a/com.unity.mobile.notifications/CHANGELOG.md
+++ b/com.unity.mobile.notifications/CHANGELOG.md
@@ -8,6 +8,7 @@ All notable changes to this package will be documented in this file.
 
 ### Fixes:
 - [iOS] - Do not copy Android plugins to Xcode project.
+- [iOS] - Calendar trigger will use local or UTC time depending on which one is set on trigger.
 
 ### Changes & Improvements:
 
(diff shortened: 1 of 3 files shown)


guobyt (approved):
Sufficiently tested by the developer. Additionally, tested whether scheduled Calendar notifications arrive on the iOS Simulator and on the physical device (iPhone 12 Pro, OS: 16.2). Also, checked on the device to see if scheduled Calendar notifications arrive when the local auto time is changed by a few minutes after the notification is scheduled.

PR #305, opened 2023-12-23
Fix crash with image attached to notification

allan-o3h:
Using a notification service extension as per: https://firebase.google.com/docs/cloud-messaging/ios/send-image we observed crashes when receiving a notification with the app opened
This change resolves the crash issue for us

### DIFF
--- a/com.unity.mobile.notifications/Runtime/iOS/Plugins/UnityNotificationData.m
+++ b/com.unity.mobile.notifications/Runtime/iOS/Plugins/UnityNotificationData.m
@@ -298,7 +298,7 @@ void _ReadNSDictionary(void* csDict, void* nsDict, void (*callback)(void* csDcit
 
 void _ReadAttachmentsNSArray(void* csList, void* nsArray, void (*callback)(void*, const char*, const char*))
 {
-    NSArray<UNNotificationAttachment*>* attachments = (__bridge_transfer NSArray<UNNotificationAttachment*>*)nsArray;
+    NSArray<UNNotificationAttachment*>* attachments = (__bridge NSArray<UNNotificationAttachment*>*)nsArray;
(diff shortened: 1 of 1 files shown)


com.unity.mobile.notifications/Runtime/iOS/Plugins/UnityNotificationData.m
```diff
@@ -298,7 +298,7 @@ void _ReadNSDictionary(void* csDict, void* nsDict, void (*callback)(void* csDcit
 
 void _ReadAttachmentsNSArray(void* csList, void* nsArray, void (*callback)(void*, const char*, const char*))
 {
-    NSArray<UNNotificationAttachment*>* attachments = (__bridge_transfer NSArray<UNNotificationAttachment*>*)nsArray;
+    NSArray<UNNotificationAttachment*>* attachments = (__bridge NSArray<UNNotificationAttachment*>*)nsArray;
```
aurimasc, line 301: This change is not correct, it causes a memory leak.
Is nsArray null when it crashes?

aurimasc (changes requested):

com.unity.mobile.notifications/Runtime/iOS/Plugins/UnityNotificationData.m
allan-o3h: nsArray is not null - this is being triggered twice for each notification (leading to a double free).

While the app is in the foreground we are listening for remote notifications (so that they don't trigger) - and also using local notifications.  This is causing UnityNotificationManager::userNotificationCenter:willPresentNotification to trigger both iOSNotificationsWrapper.NotificationReceived and iOSNotificationsWrapper.RemoteNotificationReceived causing this function to be executed twice with the same array.

I believe the proper solution is to make this change here and properly clean up the attachments in freeiOSNotificationData.

aurimasc (approved):

PR #307, opened 2024-01-08
Remove auto-added stuff from the manifest

aurimasc:
https://jira.unity3d.com/browse/MNB-76
The androidlib plugin has a manifest file that contains base stuff. Then post process script adds additional things to it based on settings. By accident after creating androidlib plugin manifest ended up having too many things, so even if you don't set rescheduling after boot or exact scheduling in the settings, these settings are still in the manifest.

Adding stuff to manifest is tested by automation. Only remove things that shouldn't be there in the first place. Exact scheduling metadata item needs to be in the manifest, as for legacy reasons code defaults to 1 otherwise.
Tested manually by playing with settings that the stuff removed in this PR gets auto-added if requested.


guobyt (approved):
Developer tested enough, no additional QA pass is needed

PR #311, opened 2024-01-23
Change repeating notification setup in test project

aurimasc:
https://jira.unity3d.com/browse/MNB-72
"Repeatable notification" has very few fields setup and had weird bahavior. On iPhone X with iOS 16.7.4 it would not arrive when app is in background, for example.
Did two changes to it: added title and changed presentation from Alert (now deprecated by Apple) to Banner. Now it works.
No changes in package code, no other changes in test project.

QA: test if this fixes the issue on device where the issue was reproduced.


guobyt (approved):
Verified that the issue no longer reproduces on the VLNQA00604 - iPhone 15 Pro Max (MU773PX/A), CPU: Apple A17 Pro, OS: 17.0.2 device, repeatable notifications now arrive with sound.

PR #312, opened 2024-02-05
Android notification status after app restart

aurimasc:
https://jira.unity3d.com/browse/MNB-78
Fix incorrect notification status being reported on app restart. When app starts, we load all scheduled notifications to a cached list, but there was no check for expiration, so if notification arrived and was dismissed with app killed and restarted afterwards, we would report notification as scheduled afterwards.

Existing test was updated to cover this case. Most basic scenarios also covered by existing tests. Also tried the repro project from the original users report that the issue no longer reproduces.
QA:
- Test reschedule after reboot
- Try various notification status reporting scenarios you can think of (like app killed at various points in time with regard to notification arrival)
- Not device or OS version specific, don't overstretch  there

### DIFF
--- a/com.unity.mobile.notifications/CHANGELOG.md
+++ b/com.unity.mobile.notifications/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable changes to this package will be documented in this file.
 
 ### Fixes:
 - [Android] - Remove excessive permissions and metadata from default manifest (auto-add them based on settings).
+- [Android] - Fixed notification status reporting after dismissing notification and re-launching the app.
 - [iOS] - Fixed a possible crash when sending push notifications with attachment (like image).
 
 ## [2.3.1] - 2023-12-11
--- a/com.unity.mobile.notifications/Runtime/Android/AndroidNotificationCenter.cs
+++ b/com.unity.mobile.notifications/Runtime/Android/AndroidNotificationCenter.cs
@@ -35,7 +35,7 @@ namespace Unity.Notifications.Android
         Scheduled = 1,
 
         /// <summary>
-        /// A notification with a specified id was already delivered.
+        /// A notification with a specified id was already delivered (showing in status bar).
         /// </summary>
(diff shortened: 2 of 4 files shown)


guobyt (approved):
For testing I used modified [test project](https://editorbugs.unity3d.com/api/v1.0/attachments/f3d806e5-d6da-4ff7-8c6b-6caca9694c73/Test.zip) from the ticket. Modified notification arrival time - changed to 60s. For Scheduled After Restart changed notification arrival time to 120s and enabled Reschedule on Device Restart setting in Mobile Notifications settings. Tested these scenarios:

1. Sending notification, terminating the app, opening app from the notification. Result after pressing status - `[Log] Unknown`
2. Sending notification, terminating the app, opening app normally. Result after pressing status - `[Log] Delivered`
3. Sending notification, putting app to the background, opening app from the notification. Result after pressing status - `[Log] Unknown`
4. Sending notification, putting app to the background, opening app normally. Result after pressing status - `[Log] Delivered`
5. Sending notification, terminating app after 50s, opening app from the notification. Result after pressing status - `[Log] Unknown`
6. Sending notification, terminating app after 50s, opening app normally. Result after pressing status - `[Log] Delivered`
7. Sending notification, restarting the device, opening app normally after notification arrives. Result after pressing status - `[Log] Delivered`
8. Sending notification, restarting the device, opening app from the notification after it arrives. Result after pressing status - `[Log] Unknown`
9. Sending notification, putting app to the background, dismissing arrived notification, opening app normally. Result after pressing status - `[Log] Unknown`
10. Sending notification, terminating the app, dismissing arrived notification, opening app normally. Result after pressing status - `[Log] Unknown`

Everything works as expected, additionally documentation was improved, to clarify Delivered status meaning.

Tested device:
- VLNQA00090 - Samsung Galaxy S8 (USA) (SM-G950U), CPU: Snapdragon 835 MSM8998, GPU: Adreno 540, OS: 7.0

PR #317, opened 2024-04-24
More reliable open with notification detection

aurimasc:
https://jira.unity3d.com/browse/PLAT-8970
Currently retrieving notification on app open is not reliable on iOS. Because of asynchronous notification delivery, when app is launched by tapping a notification, on some devices it takes a frame or two to get the notification. This can cause problems, since one may want to open different scene based on whether app was opened normally or via notification. Until now recommendation to users was to wait for a couple of frames.
This PR introuduces to API: iOSNotificationCenter.QueryLastRespondedNotification() which returns an operation that can be returned from coroutine to wait until Unity determines if app was launched via notification or not. We detect if how app was launched by looking into apps launch options.
Unified API also got the same counterpart API, on Android it finishes right away.

Testing status:
- Only impacts iOS and Unified APIs, direct Android API is not affected;
- Only impacts retrieving responded notification & actions;
- No direct dependency on device or OS version, however, the issue being solved is dependent on that (on some devices notification can be retrieved on very first frame, on others there is a small delay);
- Main test project has been updated to use new APIs in both iOS specific and in Unified variant (the separate scene);
- Developer testing:
  - Used Main testing project, both platform specific and Unified;
  - Devices used: iPhone 12 Pro (17.4), Pixel 5 (13)
  - Scenarios: with app not running or being in background open app normally, by tapping a notification, by using notification action (actions are iOS only); should show correct info about notification usage to open app (Unified scene shows notification if it was used, but only Debug.Log in case app was opened normally).
- No specific testing request for QA


com.unity.mobile.notifications/Runtime/Unified/QueryLastRespondedNotificationOp.cs
```diff
@@ -0,0 +1,114 @@
...
+    {
+        /// <summary>
+        /// Operation is ongoing, wait for next frame and check again.
+        /// </summary>
+        Pending,
+
+        /// <summary>
+        /// Operation is complete, app was launched normally, no notification was tapped.
+        /// </summary>
+        NoRespondedNotification,
```
todi1856, line 22: I assume this means "Not responded to notification" ? a bit hard to decipher the value meaning

com.unity.mobile.notifications/Runtime/iOS/QueryLastRespondedNotificationOp.cs
```diff
@@ -0,0 +1,135 @@
+using System;
+using UnityEngine;
+
+namespace Unity.Notifications.iOS
+{
+    /// <summary>
+    /// The state of the query for last responded notification.
+    /// Returned by <see cref="QueryLastRespondedNotificationOp.State"/>
+    /// </summary>
+    public enum QueryLastRespondedNotificationState
```
todi1856, line 10: Is this intentional that exactly the same enum is in com.unity.mobile.notifications/Runtime/Unified/QueryLastRespondedNotificationOp.cs ?  and both are even public.

com.unity.mobile.notifications/Runtime/iOS/QueryLastRespondedNotificationOp.cs
```diff
@@ -0,0 +1,135 @@
...
+        /// </summary>
+        HaveRespondedNotification,
+    }
+
+    /// <summary>
+    /// An operation for retrieving notification used to open the app.
+    /// When app is not running, app launches first and then notification is delivered. There may be a delay until notification is delivered.
+    /// This operation may finish immediately or it may require a few frames to pass. You can return it from coroutine to wait until completion.
+    /// </summary>
+    public class QueryLastRespondedNotificationOp
```
todi1856, line 33: same for this class - it's duplicating, just different namespace

com.unity.mobile.notifications/Runtime/iOS/QueryLastRespondedNotificationOp.cs
aurimasc: One is iOS specific API, other works on both Android and iOS. The later is in assembly that depends on former. Same for the other.
Unified APIs wrap the common parts of Android and iOS APIs (leaves specific things out, but if you only need basic stuff, it makes writing script much much simpler).

todi1856: Shouldn't then iOS part derive from unified part (Unity.Notifications.iOS.QueryLastRespondedNotificationOp derive from  Unity.Notifications.QueryLastRespondedNotificationOp)  ? and expand stuff where needed. 

Common sense say, we shouldn't duplicate code if possible

aurimasc: It can't do that now because the dependency is the opposite.

todi1856: Maybe something to refactor for the future? maybe at least I would be possible to not duplicate the enum?

todi1856 (approved):

com.unity.mobile.notifications/Runtime/iOS/QueryLastRespondedNotificationOp.cs
aurimasc: Yes, we probably could have one more assembly for common stuff. But that requires structural changes. Right now I'd like to not do them given that we will likely have to do them soon anyway, so one break rather than two.

todi1856: Maybe worth creating JIRA items for next major version bump?

com.unity.mobile.notifications/Documentation~/Unified.md
```diff
@@ -115,14 +115,18 @@ To update an existing notification, schedule a notification with the same ID. Th
...
+    yield return query;
+    if (query.State == QueryLastRespondedNotificationState.HaveRespondedNotification)
+    {
+        var lastNotification = query.Notification;
+    }
 }
 ```
 
-This also works if you send notification to the background and tap the notification to bring it back to foreground. You can query the properties of returned notification.
+This also works if you send notification to the background and tap the notification to bring it back to foreground. You can query the properties of returned notification. The operation can be completed right away, but it is not guaranteed. Operation can be returned from coroutine to wait until it is complete, in case it was complete right away, it will continue execution on next iteration.
```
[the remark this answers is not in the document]
anjalin14: I might not have understood this completely. What exactly 'operation' means here? Does it mean the action of async querying to continue until Unity figures out whether the app is launched thru tapping notification? Is it kind of an object?

com.unity.mobile.notifications/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to this package will be documented in this file.
 
+## [Unreleased]
+
+- [iOS] New, more reliable APIs (QueryLastRespondedNotification) for querying if app was opened by tapping notification.
```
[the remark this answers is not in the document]
aurimasc: Sounds like this changes a meaning a bit. If you launch app normally, there is no notification to retrieve, you can only retrieve it if app was launched by tapping the notification.

com.unity.mobile.notifications/Documentation~/Unified.md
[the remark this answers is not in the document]
aurimasc: Spotted a mistake in this one: it should be "This also works if you send application to the background" (app, not notification).
Operation here is an object returned from the api. You use that object to track if query has been completed and receive the results.

anjalin14, line 129:
```suggestion
This also works if you send the application to the background and tap the notification to bring it back to the foreground. You can query the properties of returned notification. Using QueryLastRespondedNotification API, you can track your query’s status and the notification details within the operation that the API returns as an object. You might receive the notification details immediately in the operation but it is not guaranteed. You can use the operation in a coroutine to wait until the query is complete. In case it completes immediately, the coroutine continues its execution on the next iteration.
```

com.unity.mobile.notifications/CHANGELOG.md
anjalin14, line 7:
```suggestion
- [iOS] Added a new API (QueryLastRespondedNotification) for retrieving notification details when an application is launched by tapping the notification.
```

anjalin14 (commented):
Thanks for the explanation. Adding new suggestions here.

anjalin14 (approved):

guobyt (approved):
Additionally, the following scenarios were checked:

With the Main project, MainScene:
1. Opening app normally when app is terminated
2. Opening app from notification when app is terminated
3. Opening app normally when app is put to the background
4. Opening app from notification when app is put to the background
5. Opening app using notification action when app is put to the background
6. Opening app using notification action when app is terminated

With the Main project, UnifiedAPIs:
1. Opening app normally when app is terminated
2. Opening app from notification when app is terminated
3. Opening app normally when app is put to the background
4. Opening app from notification when app is put to the background

The MainScene scene was only tested with iOS while the UnifiedAPIs scene was tested with both Android and iOS. During testing, it was found that with some scenarios, logs were not printed correctly. Aurimas resolved this issue, and now I cannot detect any additional problems.

For testing used devices:

- VLNQA00604 - iPhone 15 Pro Max (MU773PX/A), CPU: Apple A17 Pro, OS: 17.0.2
- VLNQA00352 - iPhone 12 Pro (MGML3ET/A), CPU: Apple A14 Bionic, GPU: Apple designed, OS: 16.2
- VLNQA00234 - iPhone 7 (MN8Y2ET/A), CPU: Apple A10 Fusion, GPU: PowerVR GT7600 Plus, OS: 15.7.6
- VLNQA00258 - iPhone Xs Max (MT502ET/A), CPU: Apple A12 Bionic, GPU: Apple designed, OS: 14.4
- VLNQA00498 - Google Pixel 7 (Pixel 7), CPU: Google Tensor G2, GPU: Mali-G710, OS: 13
- VLNQA00090 - Samsung Galaxy S8 (USA) (SM-G950U), CPU: Snapdragon 835 MSM8998, GPU: Adreno 540, OS: 7.0

PR #324, opened 2024-07-26
Minor iOS cleanup

aurimasc:
Notifications did have UnityAppController expansion that had no use. This has been removed along with header file.
.mm file was renamed to match the class it contains. Also we no longer create and instance of UnityNotificationLifeCycleManager as we don't use it in any way. Only hook to lifecycle notifications when framework is loaded.
The entire PR is a simple reduction of code. The only affected area is the initialization part.
Tested manually on iPhone 12 Pro Max that local notifications still work.
Tested on simulator that push notifications are still received correctly.
iOS only, not specific to version and very isolated (initialization part that happens automatically on app start).


Pavel-Unity (approved):

guobyt (approved):
No need for additional QA pass, Aurimas tests are enough.

PR #327, opened 2024-10-07
Reboot receiver entirely in the manifest

aurimasc:
https://jira.unity3d.com/browse/MNB-80
Android manifest for notifications always has a disabled receiver for rescheduling notifications after reboot. This receiver then is enabled by Java code depending on setting, value of which is also in the manifest as meta-data.
While receiver can be created by code, they are then bound to context lifecycle, as such it is better to have this receiver in the [manifest](https://developer.android.com/develop/background-work/background-tasks/broadcasts#android-14).
This PR removes the Java code to enable the receiver and moves the logic to post processing: we will only add receiver to manifest if user has enabled reschedule on boot setting.

Developer testing:
- tested that manifest items are added as requested in Unity 2022.3 and 6 (different code path before 6)
- tested that after reboot notification still arives (Pixel 5, Android 13)

Additional testing required:
- whether notifications work after reboot on more devices and Android versions; when in doubt - check without these changes

### DIFF
--- a/com.unity.mobile.notifications/Editor/AndroidNotificationPostProcessor.cs
+++ b/com.unity.mobile.notifications/Editor/AndroidNotificationPostProcessor.cs
@@ -177,11 +177,6 @@ namespace Unity.Notifications
 <manifest xmlns:android=""http://schemas.android.com/apk/res/android"" package=""com.unity.androidnotifications"">
   <application>
     <receiver android:name=""com.unity.androidnotifications.UnityNotificationManager"" android:exported=""false"" />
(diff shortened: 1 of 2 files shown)


IGuscin (approved):
Verified that Receiver manifest entry is added only when "Reschedule on Device Restart" setting is checked.

Tested that notifications properly arrive after restart.
Tested on: 
Galaxy Z Fold 6 Android 14
Motorola Nexus 6 Android 7.1.1
Pixel 6 Pro Android 14
LG V30 Android 8
Galaxy s23+ Android 13

Unity 22.3 and 6

todi1856 (approved):

PR #335, opened 2025-05-09
Add 6.1 to tested versions

aurimasc:
Unity 6.1 is current last stable release, should be tested as well.


todi1856 (approved):
Shouldn't you switch to Wrench, #rm-packageworks.

I already switched for logcat package.

PR #336, opened 2025-05-23
Integrate Wrench

aurimasc:
Partial integration of Wrench.
Package tests still use manual setup, don't want to overblow things with a single PR.


Tools/CI/regenerate.sh
```diff
@@ -0,0 +1,2 @@
+cd $(dirname "$0")/../../
+dotnet run --project Tools\CI\UnityMobileNotifications.Cookbook.csproj
```
todi1856, line 2: Shouldn't UnityMobileNotifications.Cookbook.csproj be added to repo ?

aurimasc: Good catch.

PR #351, opened 2025-06-17
Last intent docs

aurimasc:
https://jira.unity3d.com/browse/MNB-89
Update documentation on when last notification is available on Android.

### DIFF
--- a/com.unity.mobile.notifications/Documentation~/Android.md
+++ b/com.unity.mobile.notifications/Documentation~/Android.md
@@ -200,6 +200,6 @@ if (notificationIntentData != null)
 }
 ```
 
-If app is opened via notification and then sent to background, bringing it back to foreground from recent app list is considered resume, so app is still opened by tapping notification. If app receives multiple notifications, they can get grouped by the Operating System. In such case no notification will be available even if app was opened by tapping such grouped notification.
+If the user opens the app via notification, sends it to the background, then reopens it from the recent app list, the Android OS resumes the app instead of restarting it completely. If the app receives multiple notifications, the OS might group them. In this case, when the user taps such grouped notification, no individual notification data will be available.
 
 If the app is opened in any other way, [AndroidNotificationCenter.GetLastNotificationIntent](../api/Unity.Notifications.Android.AndroidNotificationCenter.html#Unity_Notifications_Android_AndroidNotificationCenter_GetLastNotificationIntent) returns null.
--- a/com.unity.mobile.notifications/Runtime/Android/AndroidNotificationCenter.cs
+++ b/com.unity.mobile.notifications/Runtime/Android/AndroidNotificationCenter.cs
@@ -1100,10 +1100,10 @@ namespace Unity.Notifications.Android
 
         /// <summary>
         /// Allows retrieving the notification used to open the app.
-        /// Unity sets notifications up to open app when tapped and put notification into the Intent.
+        /// Unity sets up notifications to open the app when the user taps on them and adds notification data into the Intent.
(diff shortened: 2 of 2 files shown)


com.unity.mobile.notifications/Documentation~/Android.md
```diff
@@ -200,4 +200,6 @@ if (notificationIntentData != null)
 }
 ```
 
+If app is opened via notification and then sent to background, bringing it back to foreground from recent app list is considered resume, so app is still opened by tapping notification. If app receives multiple notifications, they can get grouped by the Operating System. In such case no notification will be available even if app was opened by tapping such grouped notification.
```
anjalin14, line 203: Does it mean, user taps the notification to open the app but may be minimises it. Then again opens it through recent app list. So the OS thinks that it is resuming an app that was opened by the tapping notification

com.unity.mobile.notifications/Runtime/Android/AndroidNotificationCenter.cs
```diff
@@ -1099,7 +1099,11 @@ public static NotificationStatus CheckScheduledNotificationStatus(int id)
         }
 
         /// <summary>
-        /// Allows retrieving the notification used to open the app. You can save arbitrary string data in the 'AndroidNotification.IntentData' field.
+        /// Allows retrieving the notification used to open the app.
+        /// Unity sets notifications up to open app when tapped and put notification into the Intent.
+        /// This method checks the last Intent for the current activity and reconstructs notification if one is found.
+        /// Note, that bringing app from background to foreground from recent list is resume, not open, hence the Intent does not change and notification will remain present if it was there before.
```
anjalin14, line 1105:
```suggestion
        /// Note that when the user brings the app from the background to the foreground through the recent app list, the system resumes the app instead of opening it. Hence, the Intent does not change and notification data will be available if it existed before.
```

com.unity.mobile.notifications/Runtime/Android/AndroidNotificationCenter.cs
```diff
@@ -1099,7 +1099,11 @@ public static NotificationStatus CheckScheduledNotificationStatus(int id)
...
+        /// If app receives many notifications, they get grouped by OS. Tapping such grouped notification will open the app, but no notification will be returned by this method, as app in this case was not opened by a specific notification.
```
anjalin14, line 1106:
```suggestion
        /// If the app receives multiple notifications, the OS groups them. Tapping such grouped notification opens the app, but this method doesn't return any notification data. This is because the app didn't open from a specific notification.
```

com.unity.mobile.notifications/Runtime/Android/AndroidNotificationCenter.cs
```diff
@@ -1099,7 +1099,11 @@ public static NotificationStatus CheckScheduledNotificationStatus(int id)
         }
 
         /// <summary>
-        /// Allows retrieving the notification used to open the app. You can save arbitrary string data in the 'AndroidNotification.IntentData' field.
+        /// Allows retrieving the notification used to open the app.
+        /// Unity sets notifications up to open app when tapped and put notification into the Intent.
```
anjalin14, line 1103:
```suggestion
        /// Unity sets up notifications to open the app when the user taps on them and adds notification data into the Intent.
```

anjalin14 (approved):

com.unity.mobile.notifications/Documentation~/Android.md
anjalin14: Thank you

PR #365, opened 2025-08-28
Update test versions

aurimasc:
Unity 6.1 is no longer supported and 6.3 branched off from trunk.


todi1856 (approved):
Don't think you need my review for such PRs

PR #445, opened 2026-07-03
Swift trampoline support

aurimasc:
https://jira.unity3d.com/browse/PLAT-21036
Make mobile notifications package work with Swift trampoline.
Changes:
- fix subscribing to OS notifications, which are between across trampolines
- fix device token retrieval (userInfo difference between trampolines)
- fix an issue with safe area handling and notification retrieval in unpause, issue with event timing in different trampones here
Testing:
- tested both trampolines using trunk
- simple notifications are sent in both
- device token is retrieved in both
- automation currently only tests Objective-C, adding it for Swift will be done as a separate task

### DIFF
--- a/TestProjects/Main/Assets/Scripts/Helpers/SafeArea.cs
+++ b/TestProjects/Main/Assets/Scripts/Helpers/SafeArea.cs
@@ -33,6 +33,9 @@ namespace Unity.Notifications.Tests.Sample
 
         void ApplySafeArea(Rect r)
         {
+            // Very early in the load process screen size may not be determined yet
+            if (Screen.width == 0 || Screen.height == 0)
+                return;
             LastSafeArea = r;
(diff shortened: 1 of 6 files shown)


rytis-buz (approved):

guobyt (approved):
Additionally tested:

- Adding the push notifications capability and requesting permission shows `RegisteredForRemoteNotifications: True`
- Removing the push notifications capability and requesting permission shows `RegisteredForRemoteNotifications: False`
- DeviceToken is received
- Push notifications can be unregistered and registered again
- Notifications arrive when the app is put in the background
- Push notifications also arrive (checked on simulator only)

All checks were done on both the Swift and Objective-C trampolines on Unity 6000.5.3f1.
Everything works as expected. Thank you for the change!

PR #447, opened 2026-07-08
Android cleanup after min spec bump

aurimasc:
https://jira.unity3d.com/browse/PLAT-19434
Android code cleanup after min spec bump. Package now is compatible with Unity 6.0, which requires Android 6 (M, API level 23), so all the code paths for the older versions can go away.
Some unrelated changes: Rider warning cleaned up.
There is a single functional change: the code cleaning up old notifications was not correct - did not modify set passed as parameter. This change does not have user-visible effect, we simply kept information on notifications longer then we should have, past their expiration time.

Tested on Pixel 5 with Android 13:
- notifications work, tapping one opens the app
- cancel notification works
- update notification works correctly
- killing app or restarting device, scheduled notification arrives as expected

*Please carefully review the code changes so that no condition was flipped anywhere.*

### DIFF
--- a/com.unity.mobile.notifications/Documentation~/Android.md
+++ b/com.unity.mobile.notifications/Documentation~/Android.md
@@ -34,8 +34,6 @@ On devices that use Android versions prior to 8.0, this package emulates the sam
 
 ## Schedule notifications at exact time
 
-Before Android 6.0 notifications can be scheduled only at approximate time.
-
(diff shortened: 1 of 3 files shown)


todi1856 (approved):
nothing in CHANGELOG ?

aurimasc:
> nothing in CHANGELOG ?

There are no user worthy notes here. Min spec bump is already in there, changes are just code cleanup.

todi1856:
> > nothing in CHANGELOG ?
> 
> There are no user worthy notes here. Min spec bump is already in there, changes are just code cleanup.

You've changed minSdk in build.gradle from 22 to 23, I don't think that's present in changelog

aurimasc:
> You've changed minSdk in build.gradle from 22 to 23, I don't think that's present in changelog

This is what Unity 6 requires.

IGuscin (approved):
Did a small sanity check, looks good, didn't notice any issues.  
App builds fine, notifications do arrive, scheduled notifications arrive after app/device restart.

PR #448, opened 2026-07-13
Swift trampoline fixes

aurimasc:
https://jira.unity3d.com/browse/UUM-147194

Fix issues with Swift trampoline support:
- Preprocessor.h file is in a different location, so patching of it was broken
- defines were not added for Swift language, only Objective-C
- there is a difference between Swift trampoline in trunk and previous versions (UnityAPI not a thing in older); solved the issue by having swift source file to access Swift APIs

Other changes, not specific to Swift:
- Add 6.6 to tested versions
- Remove ancient check for targetting iOS 10 or later, min spec guarantees 13+
- entitlement files was not correctly added if time sensitive notifications were not enabled in settings
- fixed a bunch of trivial compiler warnings

Testing:
- Tested that Objective-C trampoline still builds and registers for push with device token retrieve on 6.0
- Tested that Swift trampoline builds and registers for push with token retrieve on 6.5 and trunk
- remainder should be covered by automation (Objective-C only)

### DIFF
--- a/com.unity.mobile.notifications/Editor/iOSNotificationPostProcessor.cs
+++ b/com.unity.mobile.notifications/Editor/iOSNotificationPostProcessor.cs
@@ -55,6 +55,7 @@ public class iOSNotificationPostProcessor : MonoBehaviour
     private static void PatchPBXProject(string path, bool needLocationFramework, bool addPushNotificationCapability, bool useReleaseAPSEnv, bool addTimeSensitiveEntitlement)
     {
         var pbxProjectPath = PBXProject.GetPBXProjectPath(path);
+        List<string> swiftPropertiesToAdd = new();
 
         var needsToWriteChanges = false;
 
@@ -87,6 +88,7 @@ public class iOSNotificationPostProcessor : MonoBehaviour
         if (needLocationFramework && !pbxProject.ContainsFramework(unityFrameworkTarget, "CoreLocation.framework"))
         {
             pbxProject.AddFrameworkToProject(unityFrameworkTarget, "CoreLocation.framework", false);
+            swiftPropertiesToAdd.Add("-DUNITY_USES_LOCATION");
             needsToWriteChanges = true;
         }
 
@@ -100,6 +102,7 @@ public class iOSNotificationPostProcessor : MonoBehaviour
         // Update the entitlements file.
         if (addPushNotificationCapability)
         {
+            swiftPropertiesToAdd.Add("-DUNITY_USES_REMOTE_NOTIFICATIONS");
             var capManager = new ProjectCapabilityManager(pbxProjectPath, entitlementsFileName, "Unity-iPhone");
             capManager.AddPushNotifications(!useReleaseAPSEnv);
             capManager.WriteToFile();
@@ -122,6 +125,12 @@ public class iOSNotificationPostProcessor : MonoBehaviour
                 pbxProject.AddBuildProperty(mainTarget, "CODE_SIGN_ENTITLEMENTS", entitlementsFileName);
                 needsToWriteChanges = true;
             }
+        }
+
+        if (swiftPropertiesToAdd.Count > 0)
+        {
+            pbxProject.UpdateBuildProperty(unityFrameworkTarget, "OTHER_SWIFT_FLAGS", swiftPropertiesToAdd, new string[0]);
+            needsToWriteChanges = true;
         }
 
         if (needsToWriteChanges)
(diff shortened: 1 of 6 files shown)


com.unity.mobile.notifications/Editor/iOSNotificationPostProcessor.cs
```diff
@@ -120,12 +111,22 @@ private static void PatchPBXProject(string path, bool needLocationFramework, boo
...
+            needsToWriteChanges = true;
+        }
+
+        if (swiftPropertiesToAdd.Count > 0)
+        {
+            pbxProject.UpdateBuildProperty(unityFrameworkTarget, "OTHER_SWIFT_FLAGS", swiftPropertiesToAdd, new string[0]);
+            needsToWriteChanges = true;
+        }
+
+        if (needsToWriteChanges)
```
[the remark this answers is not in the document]
aurimasc: No, it does not drop anything, it is serializing the same thing. And it seems `ProjectCapabilityManager.WriteToFile()` writes other files, but not the pbxproject, it only modifies the later.

com.unity.mobile.notifications/Runtime/iOS/Plugins/UnityMobileNotificationsBridge.swift
```diff
@@ -0,0 +1,25 @@
+#if UNITY_XCODE_PROJECT_TYPE_SWIFT
+
+import Foundation
+
+@_cdecl("UnityMobileNotifications_applicationWillFinishLaunchingName")
+func applicationWillFinishLaunchingName() -> String {
```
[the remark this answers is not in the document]
aurimasc: Corrected.

com.unity.mobile.notifications/Runtime/iOS/Plugins/UnityNotificationLifeCycleManager.mm
```diff
@@ -10,10 +10,16 @@
 #import "UnityNotificationManager.h"
 
 #if UNITY_XCODE_PROJECT_TYPE_SWIFT
-#import "UnityFramework/UnityFramework-Swift.h"
-#define kUnityWillFinishLaunchingWithOptions UnityNotifications.applicationWillFinishLaunching
-#define kUnityDidRegisterForRemoteNotificationsWithDeviceToken UnityNotifications.applicationDidRegisterForRemoteNotifications
-#define kUnityDidFailToRegisterForRemoteNotificationsWithError UnityNotifications.applicationDidFailToRegisterForRemoteNotifications
+extern "C"
+{
+    NSNotificationName UnityMobileNotifications_applicationWillFinishLaunchingName();
```
rytis-buz, line 15: Can't you #import UnityAPI or UnityFramework to see UnityNotifications interface here?

aurimasc: UnityAPI does not exist in 6.6 and 6.5 and it is the one you have to import in trunk. We don't have version handling yet, so I can make it work with older or newer, not both.

rytis-buz: I think we talked that in similar scenarios it should be possible to reference UnityFramework directly which should bring in the API as well.. I see that's not the case currently. Maybe we should fix that on the Swift trampoline side
But this PR will work anyway if we decide to make these changes, so it's fine

rytis-buz (approved):

guobyt (approved):
Overall developer testing looks sufficient, but I did some additional testing on iOS across Unity 6.5, 6.6 and 6.7 on Swift trampoline. Notification settings configuration and runtime behaviour perform as expected; no regressions found.

Testing scenarios covered include:

- Xcode project generation with push notifications, CoreLocation and time-sensitive notifications enabled, correct entitlements and capabilities under the Unity-iPhone target
- All three capabilities enabled together (push + location + time-sensitive), with none dropped
- Append (second) build into the same folder - capability and a single entitlements file survive, no duplicated Swift flags
- Preprocessor.h patched at the correct Swift path (UnityFramework/Prefix/) on Unity 6000.5+
- Remote notification registration on a real device:
  - Success path - RegisterForRemoteNotifications shows True, non-empty device token received, no crash
  - Failure path - RegisterForRemoteNotifications shows False, no crash

more details in [testing document](https://docs.google.com/document/d/1Xb_ZUDZei-1Fxv9iWQ-5E2e-XJuFkaqU5DTvjMn16zI/edit?usp=sharing)

PR #459, opened 2026-08-18
Use UnityAPI on iOS when using Swift trampoline

aurimasc:
https://jira.unity3d.com/browse/PLAT-21965
Change integration with the Swift trampoline to use Unity public API directly from Objective-C++ code without swift bridge.
The new implementation is much cleaner and safer, with a price that we lose compatibility with Unity 6.5, but that was accepted.

Tested on simulator using 6.6 and trunk that remote notifications have been correctly subscribed to and device token has been received.


xTheEc0 (approved):

guobyt (approved):
No need for an additional QA pass - developer testing is sufficient here. More detailed testing will be performed before the package release.
thank you!

PR #470, opened 2026-09-09
Swift trampoline fixes

aurimasc:
https://jira.unity3d.com/browse/UUM-151886
In latest trunk the location of Info.plist file changed in swift trampoline, which broke mobile notification post processor. The fix is trivial - check if file exists and fallback to new location.
Additionally, disable homebrew update as that is causing issues with unity tools we retrieve through it.

Testing:
In general the error was caught by automation.
Tested manually both trampolines to see that info plist entries are found on startup, permission request still works correctly and push notifications do work.

### DIFF
--- a/.yamato/upm-ci.yml
+++ b/.yamato/upm-ci.yml
@@ -79,6 +79,8 @@ test_iOS_Swift_{{ editor.version }}:
     model: M1
   variables:
     IOS_SIMULATOR_SDK: 1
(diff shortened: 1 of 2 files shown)


mindaugasv-unity (approved):

robertas-unity (approved):
Verified on 6.6 and 6.7 that Notifications work correctly.

PR #473, opened 2026-09-15
Forward port 2.5 branch and fix CI

aurimasc:
https://jira.unity3d.com/browse/PLAT-22315
Forward-port PR that landed to release/2.5 branch.
Fix CI:
- trunk dropped x64 Mac Editor, fix by using ARM64 in all branches
- change image to the arm64; same image as used by Swift (downside that we lose lower than iOS 16 coverage)
- homebrew and unity-downloader-cli setup same as used by swift

### DIFF
--- a/com.unity.mobile.notifications/Editor/iOSNotificationPostProcessor.cs
+++ b/com.unity.mobile.notifications/Editor/iOSNotificationPostProcessor.cs
@@ -131,7 +131,13 @@ public class iOSNotificationPostProcessor : MonoBehaviour
 
     private static void PatchPlist(string path, List<Unity.Notifications.NotificationSetting> settings, bool addPushNotificationCapability)
     {
-        var plistPath = path + "/Info.plist";
+        var plistPath = Path.Combine(path, "Info.plist");
+        if (!File.Exists(plistPath))
+        {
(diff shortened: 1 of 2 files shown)


com.unity.mobile.notifications/Editor/iOSNotificationPostProcessor.cs
```diff
@@ -131,7 +131,13 @@ private static void PatchPBXProject(string path, bool needLocationFramework, boo
 
     private static void PatchPlist(string path, List<Unity.Notifications.NotificationSetting> settings, bool addPushNotificationCapability)
     {
-        var plistPath = path + "/Info.plist";
+        var plistPath = Path.Combine(path, "Info.plist");
```
rytis-buz, line 134: Ideally you could read INFOPLIST_FILE property first with:
```pbxProject.GetBuildPropertyForAnyConfig(mainTarget, "INFOPLIST_FILE");```
 and then fallback to searching

rytis-buz (approved):
Left a comment, but low importance

com.unity.mobile.notifications/Editor/iOSNotificationPostProcessor.cs
aurimasc: Given that this specific change is a forward port and low importance, I'll leave it as is here.

PR #475, opened 2026-09-18
Only capture logcat on failure

aurimasc:
There have been multiple instances lately where capturing logcat gets stuck.
As a mitigation, capture logcat only if execution fails. If all tests are successful, skip it.


todi1856 (commented):
Honestly I don't think it's good idea. In case of failure, you sometimes want to compare failed logcat against succesful logcat, thus is important to have logcat even in succesful builds too.

todi1856:
p.s did you consider clearing logcat before test run, that would make the log smaller, and might be the reason why it sometimes gets stuck

aurimasc:
> Honestly I don't think it's good idea. In case of failure, you sometimes want to compare failed logcat against succesful logcat, thus is important to have logcat even in succesful builds too.

The problem is that the fail rate is unacceptably high. Something like every other PR has at least one failure due to this.
During last 10 days I found 4 failures for this reason, none of the PRs had any Android changes:
https://yamato.ds.unity3d.com/job/75460576/logs/execution
https://yamato.ds.unity3d.com/job/75511999/logs/execution
https://yamato.ds.unity3d.com/job/75029231/logs/execution
https://yamato.ds.unity3d.com/job/74990540/logs/execution

todi1856:
I cannot stop you from pushing such change, but in my opinion this is wrong, at the very least try clearing the logcat before test run, maybe it's simple as that for fixing this

aurimasc:
> I cannot stop you from pushing such change, but in my opinion this is wrong, at the very least try clearing the logcat before test run, maybe it's simple as that for fixing this

Made it: https://github.com/Unity-Technologies/com.unity.mobile.notifications/pull/477

I'll put this one on pause. Let's see if things get better.

aurimasc:
For record, it happened again: https://yamato.ds.unity3d.com/job/76008742/logs/execution?q=logcat

todi1856:
> For record, it happened again: https://yamato.ds.unity3d.com/job/76008742/logs/execution?q=logcat

Can you try doing something like this:
```
adb logcat -d -T 1000
```

that will dump last 1000 messages and quit

PR #477, opened 2026-09-21
Clear logcat before running tests

aurimasc:
Clear logcat buffers before running tests. Hopefully this will reduce the fail rate when gathering logs at the end.


todi1856 (approved):
Didn't know about
```
-b all
```

it looks like in unity/unity we only pass -c when clearing logcat, -b all sounds like a good addition.

aurimasc:
Credits to StackOverflow :)

